PR #251, opened 2015-02-17
Quick patch to fix #245

### LINKED ISSUE #245: Release 0.7.1 does not work with iojs
The most recent release on npm, 0.7.1, does not work with iojs. It throws an error about closeSocket being undefined on lib/client/client.js:111.

This error seems to be resolved in the master branch.

Do you have an ETA on the next release?

Thank you

AlexCline:
Seems like in newer versions of nodejs and iojs, tls.socket does emit socket close events. We don't need to listen for the net.socket events, so we can just assign closeSocket to socket.

### DIFF
--- a/lib/client/client.js
+++ b/lib/client/client.js
@@ -105,7 +105,7 @@ function setupSocket(socket, opts) {
 
   // Since tls.socket doesn't emit 'close' events, we must register to receive
   // them on net.socket instead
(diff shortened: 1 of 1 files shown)


pfmooney:
This would break functionality in node <0.12.
I'll work on an appropriate fix along with tests to ensure proper behavior going forward.

jsumners:

when opening a new issue.

PR #255, opened 2015-03-11
Expose the Client as a module-level export

tyscorp:
This is so one can modify the client's prototype (for reasons such as promisification)

### DIFF
--- a/lib/index.js
+++ b/lib/index.js
@@ -3,6 +3,7 @@
 var Logger = require('bunyan');
 
 var client = require('./client');
+var Client = require('./client/client');
 var Attribute = require('./attribute');
 var Change = require('./change');
 var Protocol = require('./protocol');
@@ -21,6 +22,7 @@ var url = require('./url');
 ///--- API
 
 module.exports = {
+  Client: Client,
   createClient: client.createClient,
 
   Server: Server,


chrisbroome:
:+1: I need this as well for promisification.

chrisbroome:
@tyscorp `ClientPool` should probably also need to be exposed for promisification as well since `createClient` either returns a `Client` or `ClientPool` depending on the options passed in.

tyscorp:
Maybe I'm missing something, but where is ClientPool? I'm not seeing any code referencing it anywhere.

tyscorp:
Rebased.

chrisbroome:
So that's weird. I'm not seeing it in the master branch on Github. In my npm installed version, `^0.7.1`, I'm seeing it at `client/pool`.

chrisbroome:
Indeed, it's in the tagged commit here: https://github.com/mcavage/node-ldapjs/blob/v0.7.1/lib/client/pool.js

chrisbroome:
OK, so apparently the internals of client connections underwent a major overhaul in commit 72bfb9b0f749832b2310c5f5a6824c458dcda760, so `ClientPool` is no longer a thing in `master`

felixfbecker:
I also opened an issue suggesting to add native promise support, but this would do the same with Promise.denodeify()

pfmooney:
Included as of bdbff63bc02d430e87bff74e30ee03a1eade58f3

felixfbecker:
It seems like bdbff63 hasn't made it into 1.0?

jsumners:

when opening a new issue.

PR #260, opened 2015-04-14
permit Anonymous Authentication Mechanism

### LINKED ISSUE #254: no support for empty credentials
hi. a bit new to poking around with ldap, but I think I found a problem. Trying to leave credentials blank (anonymous usage)
but here:
https://github.com/mcavage/node-ldapjs/blob/master/lib/client/client.js#L468

`assert.string(credentials, 'credentials');`

but a few lines down, there is support for credentials being falsy
https://github.com/mcavage/node-ldapjs/blob/master/lib/client/client.js#L480

```
var req = new BindRequest({
    name: name || '',
    authentication: 'Simple',
    credentials: credentials || '',
    controls: controls
  });
```

So before i do a pr removing the assert, wanted to check with you.

msimerson:
As documented here: http://tools.ietf.org/html/rfc4513#section-5.1.1

closes #254

### DIFF
--- a/lib/client/client.js
+++ b/lib/client/client.js
@@ -465,7 +465,7 @@ Client.prototype.bind = function bind(name,
                                       _bypass) {
   if (typeof (name) !== 'string' && !(name instanceof dn.DN))
     throw new TypeError('name (string) required');
-  assert.string(credentials, 'credentials');
+  assert.optionalString(credentials, 'credentials');
   if (typeof (controls) === 'function') {
     callback = controls;
     controls = [];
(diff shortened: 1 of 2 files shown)


pfmooney:
In the test, you're calling `bind` with `BIND_PW`.
Is that really what you want for a passwordless bind test?

msimerson:
doh! I'll get right on that.

msimerson:
test updated, with a zero length name and zero length password.

pfmooney:
Merged in 19f2c167839fc9cfb46fdb8ee1ef5275100fd578.
Thanks

jsumners:

when opening a new issue.

PR #268, opened 2015-05-20
permit search base to be an empty string

msimerson:
which is necessary when one desires to discover the 'base' for future operations. Example:

```js
var base = '';
ldap.search(base,
    {
        scope: 'base',
        filter: '(objectclass=*)',
        attributes: [ 'defaultNamingContext' ],
    },
    function (err, res) {
        // ...
    });
```

### DIFF
--- a/lib/client/client.js
+++ b/lib/client/client.js
@@ -750,7 +750,7 @@ Client.prototype.search = function search(base,
                                           controls,
                                           callback,
                                           _bypass) {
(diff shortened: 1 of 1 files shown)


pfmooney:
I apparently had a commit languishing in my workspace to fix this.
See ec062d370c6f64e72db01494b829aed544e3e366.

jsumners:

when opening a new issue.

PR #272, opened 2015-06-05
add example client with connection error handling

msimerson:


pfmooney:
What are your thoughts on updating the example to use some of the new 'setup' event stuff?
It was intended to make client initialization easier. (And should be documented before 0.8 is cut)

msimerson:
> use some of the new 'setup' event stuff

I don't know much about this. My app was throwing exceptions when it couldn't connect to the LDAP server so I blundered about until I found where to add those event handlers for connection errors.

pfmooney:
It doesn't help that I haven't written docs for it yet.

Here's a short example of using the setup event to bind:
https://github.com/joyent/node-ufds/blob/c7aa5635871aad7a630c897be2fcf059fed2e9dd/lib/index.js#L439-L455

ldapjs will run the 'setup' handlers sequentially after establishing a TCP connection but before emitting 'connect' to the client consumer.  The idea is to simplify event handling in the face of connect/reconnect scenarios.

msimerson:
Mmm, that does look like a much nicer interface.

jsumners:

when opening a new issue.

PR #300, opened 2015-09-30
Adding the possibility to set manually the flag 'secure'

brion25:
Hi, I've been using this module for a while and it turned that I have a problem where the company I work for, they use a secure connection to access to the LDAP information but they are not using the protocol: `ldaps://` they are using the protocol: `ldap://`. So I decide to modify the package, passing manually the `secure` value where I create the client:

```javascript
var client = ldap.createClient({
    url:'ldap://www.company.org',
    secure:true
});
```

Just for the record I tried with different url combinations and just modifying the module is how I got it working

### DIFF
--- a/lib/client/client.js
+++ b/lib/client/client.js
@@ -310,7 +310,7 @@ function Client(options) {
     _url = url.parse(options.url);
   this.host = _url ? _url.hostname : undefined;
   this.port = _url ? _url.port : false;
-  this.secure = _url ? _url.secure : false;
+  this.secure = options.secure ? option.secure : _url ? _url.secure : false;
   this.url = _url;
   this.tlsOptions = options.tlsOptions;
   this.socketPath = options.socketPath || false;


pfmooney:
Using `ldaps://` is the correct way to indication an LDAP connection where TLS is used.  If your company happens to be running ldaps on port 389, then you could specify that fact via the url:
```javascript
var client = ldap.createClient({
    url: 'ldaps://www.company.org:389',
});
```

brion25:
I tried with that combination, but the server is closed on port 389, it's open on port 636 but just using TLS, if you tried to access without using TLS, the client just hangs waiting to the socket to connect.

Even If I try with this combination: `ldaps://www.company.org:636`, the client just hangs waiting for the connection.

The working combination is: `ldap://www.company.org:636` using TLS.

pfmooney:
It doesn't make sense why the URL parsing wouldn't work for this.  If you can write a test case which exposes such a bug, please raise it and we can work through the issue.

Thanks

PR #345, opened 2016-02-10
Do not call toString on buffers

sinistersig:


jsumners:

when opening a new issue.

PR #380, opened 2016-06-10
support given localAddress binding in connection creation

artic77:
Support localAddress binding if default IP is not enough.


Zaba999:
@pfmooney or @mcavage, can you support here. Is it possible to merge this change and publish new version of ldapjs? As for tests they pass on same nodejs version locally :/

Zaba999:
Any feedback please?

jsumners:

when opening a new issue.

Issue #382, opened 2016-06-15
How to gracefully close connection to the server (as a client)?


regulatre:
I'm using ldapjs as a client and I would like to gracefully close the connection when I'm done performing my query. I can't find a way to do this. Is it possible? Thanks!

nixelsolutions:
I'm in the same case, seems like I'm suffering a connection leak :(

geoffginley:
Are you guys are using SSL and getting a "ECONNRESET" which crashes node? If so I was previously having the same issue and did 2 things

1. I wrap the work in a function, I still use client.unbind() but since doing this I have not seen re-occurences.
2. I catch the exception just in case so it doesn't crash node

`process.on('uncaughtException', function (err) {
  if (err.code === "ECONNRESET") {
    console.log(err.stack);
  } else {process.exit;}
})`

I should also add I've had similar issues with TLS outside of ldapjs

nixelsolutions:
No, in my case just found it was a problem with passport ldap strategy. Unbind is working OK and actually closing connections in the right way. Passport was leaking connections cause my implementation wasn't unbinding connections.

UziTech:
The ldap client has an undocumented method called `destroy` that will close the connection and not allow reconnection

https://github.com/mcavage/node-ldapjs/blob/master/lib/client/client.js#L974

gbouthenot:
Thank you !

gbouthenot:
Thank you ! Calling ```client.destroy()``` inside the search callback closes the connection.

prajnavantha:
Is doing `client.destroy()` a right thing to do?
Won't destroying the connection and again recreating it, affect the performance.
I'm using `client.destroy()` presently but not sure of the right way. Maybe reusing the same connection is ideal...

After i tried the option `reconnect: true` i could find that once the connection is reset its getting connected again.  This may be a right thing to do but need further inputs from all...

minotaurrr:
@prajnavantha Are you getting the ECONNRESET error with the `reconnect: true` option?

prajnavantha:
No... I'm not getting an ECONNRESET after `reconnect: true`. It seems to be working fine. I'm still observing though.

jsumners:
Closing as the documentation is clarified in #393. Please open a new issue if you have further questions.

jsumners:
⚠️ This issue has been locked due to age. If you have encountered a recent
problem that seems to be covered by this issue, please open a new issue.

[...]

Issue #365, opened 2016-03-25
Importance of calling unbind


ORESoftware:
I have this code:

``` js
                adminClient = ldap.createClient(clientOptions);

                adminClient.bind(ldapConfig.ldapBindAccountDN, ldapConfig.ldapBindAccountPassword, function (err) {
                    if (err) {
                        cb(err);
                    }
                    else {
                        var searchQuery = '(sAMAccountName=' + username + ')';

                        performUserSearch(adminClient, searchQuery, 1, function (err, users) {

                            //adminClient.unbind(function ($err) { //async
                            //
                            //    if ($err) {
                            //        console.error('UNBIND ADMIN CLIENT ERROR: ', $err);
                            //    }
                            //
                            //    if (err) {
                            //        log.error(err);
                            //        cb(err);
                            //    }
                                 if (!users) {
                                    log.debug('Error: no users found at all.');
                                    cb({status: 403, error: new Error('Bad username.')});

// ..... truncated ....
```

As you can see I have commented out the unbind portion. Is it imperative to call unbind? I commented it out, and it doesn't cause any immediate problems. I am having trouble finding information on the purpose of bind/unbind. I assume unbind is important, but then why am I getting away without it? I was getting away without it _before_ I moved to pooled connections. Please enlighten me, thanks.

ORESoftware:
unfortuantely, the only info in the docs for unbind are:

```
unbind

unbind(callback)

Performs an unbind operation against the LDAP server.

Example:

client.unbind(function(err) {
  assert.ifError(err);
});
```

ORESoftware:
furthermore, if LDAP connections are stateful, does that mean having 100 connections in a pool is at all more costly that a pool of 100 stateless connections?

Xotl:
Hi @ORESoftware, i have been using this module too and yes, there's not enough documentation about the `unbinding` method. 

I use that method to close the connection, and it seems to do that because when i run my node app it keeps waiting until i call the unbind method.

**Off topic:** Also i would suggest to you instead of using `if / else` for your error handling do something like this:

```
function (err, someArgs) {
  if (err) {
    // the return will prevent the code
    // outside the if statement to run
    return cb(err);
  }

  // Code if no error here
}
```

jsumners:
Closing as the documentation is clarified in #393. Please open a new issue if you have further questions.

jsumners:
⚠️ This issue has been locked due to age. If you have encountered a recent
[...]

PR #393, opened 2016-10-16
doc: more explanation about unbind

gramakri:
Explain in more detail that `unbind` is the API to call for disconnecting.


gramakri:
This should address some of the confusion in issues including #365 and #382

jsumners:

when opening a new issue.

PR #405, opened 2016-11-22
destroy socket if auto-bind fails

pbatey:
With a client, when using bindDN and bindCredentials, if the credentials are bad then the bind fails as expected -- but the socket is left open. An explicit call to `client.destroy()` is required in order to release the socket. This is confusing since the socket would be closed after a successful operation. Closing the socket upon bind failure makes it consistent.

Also, the bind error was being reported twice. Once as an emit, and again in the callback. Because a bind failure ends with the callback being executed, an emit doesn't make sense.

The test that applies to this behavior is *auto-bind bad credentials* in `test/client.test.js` -- it continues to pass with the changes I've made, but because it calls `clt.destroy()` explicitly, it doesn't replicate the problems I was having.

### DIFF
--- a/lib/client/client.js
+++ b/lib/client/client.js
@@ -348,6 +348,9 @@ function Client(options) {
     this.on('setup', function (clt, cb) {
       clt.bind(options.bindDN, options.bindCredentials, function (err) {
         if (err) {
+          if (self._socket) {
+            self._socket.destroy()
+          }
           self.emit('error', err);
         }
         cb(err);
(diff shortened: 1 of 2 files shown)


denis-peshkov:
Guys, I have the same issue, so merge it please asap.

pbatey:
I have rebased my change, added a unit test to check that the socket is closed when bad credentials are provided, and adjusted the pull request to merge into to ldapjs:next.

jsumners:

when opening a new issue.

Issue #406, opened 2016-12-06
Wrong event names after uglify-ing the code


strebl:
I have an issue with the ldap client.

In one of my build step, the javascript code gets uglyfied. In that step, functions getting renamed etc. Normally that's not a problem, but because the event name in the client is generated off of the [message name](https://github.com/mcavage/node-ldapjs/blob/master/lib/client/client.js#L1409), a wrong event will be raised.

I think instead of getting the name from `msg.constructor.name`, getting it with `msg.type` would fix the problem.

I'll create a PR with my solution.

strebl:
🔔 Ping @pfmooney 😊

grogi:
https://github.com/joyent/node-ldapjs/pull/492

jsumners:
I am unclear why one would even need to minify/uglify this library. The intention of this library is to be used in a Node.js environment. Minification does nothing in such an environment.

jsumners:
👋

On February 22, 2023, we released [version 3](https://github.com/ldapjs/node-ldapjs/releases/tag/v3.0.0) of this library. As a result, we are closing this issue/pull request.

Please see [issue #839](https://github.com/ldapjs/node-ldapjs/issues/839) for more information, including how to proceed if you feel this closure is in error.

PR #407, opened 2016-12-06
Raise the correct event name

strebl:
This PR is in response to the issue #406 which fixes the event name for uglyfied code.

All tests returned green ✅


jsumners:
Closing in favor of #492. It seems to be more robust. If you disagree @strebl please state your case.

PR #423, opened 2017-03-08
Check for null or undefined values

martinscholz83:
Sometimes you have a fix mapping where you have to add empty values. For example
```
"sn": "Mustermann",
"givenName: "Max",
"l": [null_or_undefined_or""]
```
Currently this is not possible. This change just check if value is not null or undefined,... .


jsumners:

when opening a new issue.

Issue #329, opened 2015-12-16
Concurrent calls to client.search() failed when using "ldaps"


javefang:
I'm wondering if the functions on a client can be called concurrently? 

I have a small test application which takes a list of usernames, fetch their details from ldap using ldapjs and print them to the console. Currently I implement it in the way that it create a single ldapjs client, bind it and then call "search" for each of the usernames concurrently.

It works fine when I'm connecting with unsecured "ldap" protocol. But when switched to "ldaps", all search still returns the event emitter, but some of them will never call "end" or "error".

I'm wondering if it is safe to call another search while one is still not finished? Thanks.

Environment:
windows 7
node v0.12.9
ldapjs 1.0.0

javefang:
might have found the problem:

in lib/client/client.js, on the last few lines of function "Client.prototype._sendSocket", 

```
conn.write(message.toBer(), writeCallback);
```

If I understand correctly, this guarantees that the message (search request) is written to the socket before calling writeCallback, which will return the event emitter back to the caller. However, this might be too late in some cases. As I have noticed that the "end" event is fired before the callback of "search" is called. In which case the caller will never be able to receive that event as it hasn't got a chance to register event handler yet.

Maybe we should make sure the emitter is returned to the caller so that they have the chance to register event handler before any event arrive? My initial thought is as follow. Hopefully you could provide more help on this?

```
try {
    writeCallback();
    return conn.write(message.toBer());

  } catch (e) {
    if (timer)
      clearTimeout(timer);

    log.trace({err: e}, 'Error writing message to socket');
    emitter.emit('error', e);
  }
```

javefang:
btw, the issue I'm having only seem to affect "ldaps" protocol so far. and the above fix works for me.

pfmooney:
@javefang Could you collect trace-level logs from this situation?

Loki-Afro:
any plan in integrating this? there is also a fork with this fix only based on an old version .. https://github.com/hpi-schul-cloud/node-ldapjs/commit/3d470ab339c203560ad9816158dfb5695fff36ab

UziTech:
If someone wants to create a PR with the fix we can look at it. Make sure to add a test as well.

jsumners:
👋

On February 22, 2023, we released [version 3](https://github.com/ldapjs/node-ldapjs/releases/tag/v3.0.0) of this library. As a result, we are closing this issue/pull request.

[...]

PR #424, opened 2017-03-11
Adding patch for ldaps concurrent failures.

RenEvo:
As per issue #329 a race condition can occur that will emit events before the callback has had a chance to listen to them. We were able to replicate this about 60% of the time, causing timeout issues internally when in fact while looking at the trace logs, the search results returned.

Easiest to replicate with single result searches.

With these changes, we 100% verified that we no longer are missing events and it works as intended.

Thanks to @javefang for the fix!

### DIFF
--- a/lib/client/client.js
+++ b/lib/client/client.js
@@ -1473,12 +1473,14 @@ Client.prototype._sendSocket = function _sendSocket(message,
     log.trace('sending request %j', message.json);
 
   try {
-    return conn.write(message.toBer(), writeCallback);
+    writeCallback();
+    return conn.write(message.toBer());
   } catch (e) {
     if (timer)
       clearTimeout(timer);
 
     log.trace({err: e}, 'Error writing message to socket');
+    emitter.emit('error', e);
     return callback(e);
   }
 };


javefang:
Glad that you got this to work! Nice job, @pfmooney any chance you could take a look at this PR, please?

javefang:
Any chance we can merge this PR? Without it the ldaps:// protocol is not really usable.

cvillemure:
I have the same issue with LDAP/TLS where about 1% of the search never receive the "end" event.  This commit seems to fix the issue.

rdubigny:
Any chance to get this merged? It worked like a charm on our project witch makes hundreds of concurrent call to our LDAP server.

rdubigny:
@RenEvo can you publish your fork on npm?

RenEvo:
Since this is "technically" managed by Joyent now, would want to find out if they are going to merge this or not first.

If not, will find out if we can.

RenEvo:
@melloc thoughts?

rdubigny:
@RenEvo I will publish a fork from your repo since we need this on our project quite soon. I'll be glad to fetch your own publication if you publish it.

According to https://github.com/joyent/node-ldapjs/issues/461 , I don't expect an update on this repo.

rdubigny:
@RenEvo I just published the ldapjs-concurrency-fix package.

melloc:
@RenEvo Sorry for the delayed reply. I haven't had a chance since migrating this and several other packages over to go through all the pending merge requests and bugs, but I'd like to try to do so next week. Was there a reason you moved the writeCallback away from the .write() call? At first glance, I would expect it to not be called until the write succeeds.

RenEvo:
The emitter was not able to be wired up properly due to not having been sent to the callback yet.

#329  explains this pretty well, the fix was to call the callback so the emitter can be listened, then actually send the request across the wire. (writeCallback, then send message)

RenEvo:
This can be reproduced (in our case) by having an ldap search happen on a single connection through HTTP API over STARTTLS or TSL LDAP connection.

RenEvo:
Any progress on this @melloc ?

jsumners:

when opening a new issue.

Issue #356, opened 2016-03-11
Unhandled error event


digicasi:
This is the error I'm seeing when trying to look up a user:

2016-03-11T15:08:57.94+0100 [App/0]   ERR events.js:141
2016-03-11T15:08:57.94+0100 [App/0]   ERR       throw er; // Unhandled 'error' event
2016-03-11T15:08:57.94+0100 [App/0]   ERR       ^
2016-03-11T15:08:57.94+0100 [App/0]   ERR Error: getaddrinfo ENOTFOUND company.com:636
2016-03-11T15:08:57.94+0100 [App/0]   ERR     at errnoException (dns.js:26:10)
2016-03-11T15:08:57.94+0100 [App/0]   ERR     at GetAddrInfoReqWrap.onlookup [as
 oncomplete](dns.js:77:26)

using node v4.4.0, ldapjs 1.0.0.

I suspect that the ldap server is not accessible, but this shouldn't result in an unhandled error event, should it?
Thanks

antonfisher:
I do not think it is an good idea, but you could use empty handler for ignore errors as workaround:
```javascript
var client = ldap.createClient(opts);
client.on('error', function(){});
```
Or handle any other errors here and ignore only _ENOTFOUND_.

sebaplaza:
This bug still persists in 2018 (version 1.0.2)

What happened with #433 bugfix ?

KaiWilke:
I think this problem should be considered as a documentation bug, since the outlined examples on http://ldapjs.org/client.html do not explain how to securely setup a ldap connection...

```
// Define LDAP client connection setting

var ldap_client_object = {
	url: 'ldap://192.168.0.1:389/',
	connectTimeout: 2000,
	timeout: 5000,
};

// Create LDAP client connection 

var ldap_client = ldap.createClient( ldap_client_object );

// Register a LDAP client connection "error" handler

ldap_client.on( 'error', function( err ) {

	// The LDAP client connection has generated an error...

});

// Register a LDAP client connection "connectTimeout" handler

ldap_client.on( 'connectTimeout', function( err ) { 

	// The ldap connection attempt has been timed out...

});

// Register a LDAP client connection "connect" handler

ldap_client.on( 'connect', function( ) {

	// The ldap connection is ready to use. Place your subsequent ldapjs code here...

});
```

HTH and Cheers, Kai

tastypackets:
[...]

PR #433, opened 2017-04-28
prevent bind exceptions from crashing Node.js

mnakama:
This prevents node crashes from Client.bind() and etc.
We set a callback handler in the instance during bind()
and send errors to it.

Fixes https://github.com/mcavage/node-ldapjs/issues/356

### DIFF
--- a/lib/client/client.js
+++ b/lib/client/client.js
@@ -317,6 +317,10 @@ function Client(options) {
 
   this.log = options.log.child({clazz: 'Client'}, true);
 
+  this.on('error', function(err) {
+    self.cb(err);
+  });
+
   this.timeout = parseInt((options.timeout || 0), 10);
   this.connectTimeout = parseInt((options.connectTimeout || 0), 10);
   this.idleTimeout = parseInt((options.idleTimeout || 0), 10);
@@ -362,6 +366,17 @@ function Client(options) {
 util.inherits(Client, EventEmitter);
 module.exports = Client;
 
+/**
+ * Default handler for error callbacks when
+ * one isn't set in the instance
+ */
+Client.prototype.cb = function(err, ret) {
+  if (err) {
+    console.error('Caught exception:', err);
+  } else {
+    console.log('Unhandled output:', ret);
+  }
+};
 
 /**
  * Sends an abandon request to the LDAP server.
@@ -480,6 +495,12 @@ Client.prototype.bind = function bind(name,
     credentials: credentials || '',
     controls: controls
   });
+
+  var self = this;
+  this.cb = function(err, ret) {
+    delete self.cb;
+    callback(err, ret);
+  };
 
   return this._send(req, [errors.LDAP_SUCCESS], null, callback, _bypass);
 };


agInteg:
This fixes the bind bug where there is a problem with the LDAP server is unavailable and it results in an exception that cannot be handled in an easy way. Please merge this.

trulore:
I took the code from this PR and manually merged it into my own code base, and it fixed a critical bug we were having that crashed node several times per day.    Thank you very much for this!

mnakama:
@trulore 
Cool, I'm glad it helped you =). It's a small change, but it took me a few hours to figure out what spot needed the change. The whole error handling system in this code is done with event dispatches... which weren't getting handled gracefully. Seems hard to follow, but this is ES5 code.

@agInteg 
Seems this project needs a new maintainer. It's been half a year since I opened this PR. Though a few people have forked my code for this fix, I'm not a good candidate to take over since I don't work at the company where I implemented this anymore (and thus I currently have no use for the code).

@mcavage 
If you still check your github notifications, maybe you'll see this. Would you mind merging this PR for the community? It's very small and fixes a crucial crash bug.

zont:
Need this bugfix!

sebaplaza:
I confirm, this fix is working for #356.
Please Merge, we're having serious issues with this bug.

lib/client/client.js
```diff
@@ -362,6 +366,17 @@ function Client(options) {
 util.inherits(Client, EventEmitter);
 module.exports = Client;
 
+/**
+ * Default handler for error callbacks when
+ * one isn't set in the instance
+ */
+Client.prototype.cb = function(err, ret) {
+  if (err) {
+    console.error('Caught exception:', err);
```
jsumners:
```suggestion
    this.log.error('Caught exception:', err);
```

lib/client/client.js
```diff
@@ -362,6 +366,17 @@ function Client(options) {
...
+  } else {
+    console.log('Unhandled output:', ret);
```
jsumners:
```suggestion
    this.log.info('Unhandled output:', ret);
```

DavidIHunt:
PLEASE merge this...
It's been 2 years!

The bug it fixes is making this module awful to use. Imagine trying to call a function of a library which can asynchronously throw an error that you can't catch and which will crash your process.

UziTech:
@DavidIHunt if you want to take this on that would be great. I think this needs to be rebased on the `next` branch and a few tests added.

UziTech:
Is the issue here that the `this._send` call throws an error instead of passing back to the `callback`?

mnakama:
> Is the issue here that the `this._send` call throws an error instead of passing back to the `callback`?

If I remember correctly, there was no handler for this error that I could find, so it just threw exceptions.

UziTech:
@DavidIHunt Would this code work to catch the error?

```js
const client = ldapjs.createClient(...)
client.on('error', (err) => {
	if (err.code === 'ECONNREFUSED') {
		// handle connection refused (LDAP server not started)
	}
}).on('connectError', (err) => {
	// handle connection timeout (firewall problem)
})
client.bind(...)
```

vdechef:
> @DavidIHunt Would this code work to catch the error?
> 
> ```js
> const client = ldapjs.createClient(...)
> client.on('error', (err) => {
> 	if (err.code === 'ECONNREFUSED') {
> 		// handle connection refused (LDAP server not started)
> 	}
> }).on('connectError', (err) => {
> 	// handle connection timeout (firewall problem)
> })
> ```

This workaround does not fix the problem when the LDAP server does not answer (for example when your account is locked).

But I did try with the _next_ package (_npm install ldapjs@next_) and I was unable to reproduce the bug. So for me it seems fixed in _next_ branch.

jsumners:
Given the response https://github.com/ldapjs/node-ldapjs/pull/433#issuecomment-577263109

and the lack of response from original authors, I am closing this PR. If someone discovers it is _not_ resolved in the `next` branch, they are welcome to open a new PR targeting that branch which fixes the issue.

jsumners:

when opening a new issue.

Issue #434, opened 2017-05-03
client api, how to delete attribute


ManfredSteiner:
I want to delete an attribute, following the client api doc [ldapjs client api #modify][1] but the described method does not work.

See also [http://stackoverflow.com/questions/43759952/ldapjs-how-to-delete-attribute][2]

How I have to prepare the change object to delete an attribute?

[1]: http://ldapjs.org/client.html#modify
[2]: http://stackoverflow.com/questions/43759952/ldapjs-how-to-delete-attribute

jsumners:
Solved in #435 .

jsumners:
⚠️ This issue has been locked due to age. If you have encountered a recent
problem that seems to be covered by this issue, please open a new issue.

Please include a [minimal reproducible example](https://en.wikipedia.org/wiki/Minimal_reproducible_example)
when opening a new issue.

PR #435, opened 2017-05-03
Fix issue #434, deleting attribute

ManfredSteiner:
Now the change/modification value can be set to `undefined` or `null`, in order to indicate a non multi-valued attribute. As result `changes[0]._modification._vals` in `return this._send(req, [errors.LDAP_SUCCESS], null, callback)` (see [client.js#672][1]) is an empty array. The request is accepted by the LDAP-server (OpenLDAP) and the attribute is deleted.

To execute this operation use the following template (replace *attributeName* by the real name):

    let change = { operation : 'delete', modification: { attributeName : undefined }};
    client.modify(dn, change, (err, res) => { ... });

[1]: https://github.com/mcavage/node-ldapjs/blob/master/lib/client/client.js#L672

### DIFF
--- a/lib/change.js
+++ b/lib/change.js
@@ -82,7 +82,7 @@ Object.defineProperties(Change.prototype, {
         val[k].forEach(function (v) {
           _attr.addValue(v.toString());
         });
-      } else {
+      } else if (val[k] !== undefined && val[k] !== null) {
         _attr.addValue(val[k].toString());
       }
       this._modification = _attr;
--- a/test/issues/issue434.test.js
+++ b/test/issues/issue434.test.js
@@ -1,0 +1,26 @@
+var Logger = require('bunyan');
+
+var test = require('tape').test;
+var uuid = require('node-uuid');
+
+
+///--- Globals
+
+var ldap;
+var Change;
+
+test('issue #434', function (t) {
+    ldap = require('../../lib/index');
+    Change = ldap.Change;
+    try {
+      var change = new Change({
+        type: 'Delete',
+        modification: { cn: null }
+      });
+      t.ok(true);
+      t.end();
+    } catch (err) {
+      t.ifError(err);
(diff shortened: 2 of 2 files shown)


nhuray:
Are you considering to merge this PR ? 

This bug is annoying cause it's impossible to remove attributes from the LDAP !

ManfredSteiner:
@nhuray I have no write access to merge anything. Pull request is still open.

tastypackets:
@nhuray This repo is pretty much dead, check out LDAPts as an alternative for the client side.
https://www.npmjs.com/package/ldapts

ManfredSteiner:
This test would fail on original source and pass after modification in file client.js:  
[https://github.com/ManfredSteiner/node-ldapjs/blob/master/test/issues/issue434.test.js](https://github.com/ManfredSteiner/node-ldapjs/blob/master/test/issues/issue434.test.js)

A good location for this test would be after line 515 in file client.js:  
[https://github.com/ldapjs/node-ldapjs/blob/master/test/client.test.js#L515](https://github.com/ldapjs/node-ldapjs/blob/master/test/client.test.js#L515)

jsumners:
@ManfredSteiner thank you for coming back to work on this long neglected PR. A couple of things and we can get it merged:

1. Would you be able to rebase it to the `next` branch?
2. Move the test into the `client.test.js` file as you suggested. Just put a comment before the test that points to the issue, e.g.:

```js
// https://github.com/ldapjs/node-ldapjs/pull/435
test('can delete attributes', function (t) {
```

ManfredSteiner:
Moved to new pull request (rebased to branch next and test added).:  
See [https://github.com/ldapjs/node-ldapjs/pull/523](https://github.com/ldapjs/node-ldapjs/pull/523)

jsumners:

when opening a new issue.

Issue #439, opened 2017-06-14
Bitwise filtering is not enabled -- (UserAccountControl:1.2.840.113556.1.4.803:=2)


fuka-michal:
Im not able to filter with bitwise operators.

Filter like this: (UserAccountControl:1.2.840.113556.1.4.803:=2)

Please see https://support.microsoft.com/en-us/help/269181/how-to-query-active-directory-by-using-a-bitwise-filter

jsumners:
👋

On February 22, 2023, we released [version 3](https://github.com/ldapjs/node-ldapjs/releases/tag/v3.0.0) of this library. As a result, we are closing this issue/pull request.

Please see [issue #839](https://github.com/ldapjs/node-ldapjs/issues/839) for more information, including how to proceed if you feel this closure is in error.

PR #444, opened 2017-06-20
Fix bitwise error when using Extensible Filter

### LINKED ISSUE #430: Boolean attributes aren't retrieved when I pull all attributes from LDAP for 1 user
Boolean attributes aren't retrieved when I pull all attributes from LDAP for 1 user. Is there a bug that prevent having boolean properties? I am retrieving user data from Windows active directory.

nickzelei:
This change is fixing an error that occurs when the extensible filter tries to retrieve the value for dnAttributes. 

This is reproducible when issuing a search with this filter:

```
const extFilter = new ldapjs.ExtensibleFilter({
                attribute: "userAccountControl",
                matchType: "userAccountControl",
                rule: "1.2.840.113556.1.4.803",
                value: "2",
                dnAttributes: true
});

ldapClient.search("<base dn>", { filter: extFilter }, (err, res) => { ... });
```
This will result in the error `Expected 0x1: got 0x84`

This should fix #430 and #439

### DIFF
--- a/lib/filters/ext_filter.js
+++ b/lib/filters/ext_filter.js
@@ -39,7 +39,7 @@ ExtensibleFilter.prototype.parse = function (ber) {
       this.value = ber.readString(tag);
       break;
     case 0x84:
-      this.dnAttributes = ber.readBoolean(tag);
+      this.dnAttributes = (ber.readString(tag) | 0) !== 0;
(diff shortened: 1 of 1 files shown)


jsumners:

when opening a new issue.

PR #492, opened 2018-10-17
Client won't work after minification is done. 

grogi:
The client did not work after the uglification/minification was performed.

Replaced the manipulations on the name of the class with an expect results of said manipulations.

### DIFF
--- a/lib/client/client.js
+++ b/lib/client/client.js
@@ -1405,10 +1405,10 @@ Client.prototype._sendSocket = function _sendSocket(message,
     if (expect === 'abandon')
       return sendResult('end', null);
 
-    if (msg instanceof SearchEntry || msg instanceof SearchReference) {
-      var event = msg.constructor.name;
-      event = event[0].toLowerCase() + event.slice(1);
-      return sendResult(event, msg);
+    if (msg instanceof SearchEntry) {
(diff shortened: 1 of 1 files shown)


frostie614:
Hi @jsumners  I am new to all this but I was wondering what needs to be done to get this fix merged. I am willing to contribute with some guidance. Thank you

jsumners:
@frostie614 I basically just need to know if this change is covered by any of the unit tests. And, if not, a unit test added for it.

frostie614:
@jsumners I see there are tests in test/client.test.js which have assertions on 'searchEntry' and 'searchReference' and ensure they are okay. Is that all that needs testing? If not, what would you like covered by tests and I will check if there is a test to cover it, or implement one if there is not.

jsumners:
I missed earlier that this is targeting the `master` branch. It should also be refactored to target the `next` branch. Once that is done, you should be able to run `npm run test:cov:html` to get a report that will allow you to determine if both branches are covered by the test suite.

frostie614:
I have merged the change into the next branch.
I have run the coverage report. Both branches are covered.
Is that all? Do I need to raise a new pull request?
Thank you

jsumners:

when opening a new issue.

PR #520, opened 2019-08-15
RFC3062 server compatibility

dododedodonl:
Some extended messages, use `requestValue` not as a string, but for example a sequence.
This change makes it possible to retrieve the `requestValue` also as a buffer: `ExtendedRequest.valueBuffer`.

There is one minor change, `ExtendedRequest.value` now always returns a `string`.


jsumners:

when opening a new issue.

PR #521, opened 2019-08-25
Update ldap filter

UziTech:
- Update ldap-filter
- Convert escaped characters to Ascii hex code

ldap-filter v0.3.0 has a few [breaking changes](https://github.com/pfmooney/node-ldap-filter/blob/master/CHANGES.md#030).

Some of the tests are breaking because they don't allow special characters in an attribute name.

### DIFF
--- a/lib/filters/index.js
+++ b/lib/filters/index.js
@@ -171,9 +171,21 @@ function cloneFilter(input) {
   }
 }
 
+function escapedToHex(str) {
+  return str.replace(/\\([0-9a-f][^0-9a-f]|[0-9a-f]$|[^0-9a-f]|$)/gi, (match, p1) => {
+    if (!p1) {
+      return '\\5c';
+    }
+
+    const hexCode = p1.charCodeAt(0).toString(16);
+    const rest = p1.substring(1);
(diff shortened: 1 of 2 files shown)


UziTech:
It looks like you still need to enable the [travis builds](https://travis-ci.com/ldapjs/node-ldapjs/) on /ldapjs/node-ldapjs/

UziTech:
You should also setup appveyor for testing on windows

jsumners:
> It looks like you still need to enable the [travis builds](https://travis-ci.com/ldapjs/node-ldapjs/) on /ldapjs/node-ldapjs/

I am waiting on a reply for an application to GitHub Actions so that I can setup CI with GitHub CI. Currently just running tests manually where I need to and accepting blindly where it looks good. There's a long road to a new release...

UziTech:
> Why remove the tests?

The tests that I removed tested special characters in attribute names which is no longer supported in ldap-filter.

> How about adding a test that explicitly tests the new functionality?

There are already tests that cover special characters in the attribute values. I left those alone and made sure those are still passing.

jsumners:

when opening a new issue.

PR #523, opened 2019-08-26
Fixing issue #434 (can delete attributes)

ManfredSteiner:
Now the change/modification value can be set to `undefined` or `null`, in order to indicate a non multi-valued attribute. As result `changes[0]._modification._vals` in `return this._send(req, [errors.LDAP_SUCCESS], null, callback)` (see [client.js#672][1]) is an empty array. The request is accepted by the LDAP-server (OpenLDAP) and the attribute is deleted.

To execute this operation use the following template (replace *attributeName* by the real name):

    let change = { operation : 'delete', modification: { attributeName : undefined }};
    client.modify(dn, change, (err, res) => { ... });

[1]: https://github.com/mcavage/node-ldapjs/blob/master/lib/client/client.js#L672

### DIFF
--- a/lib/change.js
+++ b/lib/change.js
@@ -82,7 +82,7 @@ Object.defineProperties(Change.prototype, {
         val[k].forEach(function (v) {
           _attr.addValue(v.toString());
         });
-      } else {
+      } else if (val[k] !== undefined && val[k] !== null) {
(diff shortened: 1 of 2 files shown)


jsumners:

when opening a new issue.

PR #527, opened 2019-08-26
use nyc instead of istanbul

UziTech:
Update tests to use nyc instead of istanbul

nyc doesn't work in node < 6 so we wouldn't be able to run tests in older versions of node


jsumners:

when opening a new issue.

PR #536, opened 2019-08-27
Refactor tests to be independent of each other

### LINKED ISSUE #534: Fix usages of `new Buffer`
`new Buffer` has been deprecated. The instances of this in the code need to be updated to use `Buffer.from` instead.

jsumners:
This PR removes the side effect nature of the tests so that they can all be run in parallel. It also resolves #534.


jsumners:

when opening a new issue.

PR #546, opened 2019-08-28
Remove binary scripts

jsumners:
I'll open a new issue to create a new `@ldapjs/ldap-tools` package after at least a beta release of v2.0.0 is available. The binary scripts rely on a logger for tracing and I don't want to include a fully logger just for those scripts. Plus, all of the examples (in the "guide" and the "examples" documentation) do not reference them; they instead reference the standard system utility `ldapsearch`. So these binary scripts are a nice to have, but I suspect very few people are actually using them.


jsumners:

when opening a new issue.

PR #548, opened 2019-08-28
Refactor client RequestQueue into testable module

jsumners:
This is basically a step 1 in trying to make the code base a little easier to maintain and test.

Also, moves the test file glob into `.taprc` because `test/**/*.test.js` was not doing a full recursive expansion of all test directories. This is why `'test/**/*.test.js'` was necessary.

### DIFF
--- a/.taprc
+++ b/.taprc
@@ -1,1 +1,4 @@
 esm: false
+
+files:
+  - 'test/**/*.test.js'
--- a/lib/client/client.js
+++ b/lib/client/client.js
@@ -1,5 +1,7 @@
 'use strict'
 
+const requestQueueFactory = require('./request-queue')
+
 var EventEmitter = require('events').EventEmitter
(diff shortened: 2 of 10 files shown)


jsumners:
@UziTech can I get your review on this?

UziTech:
Any reason you aren't using a class? It seem weird that purge calls `this.flush` when there is no `flush` method attached to that function. I know it will have a flush method but seem like a code smell.

If it walks like a duck and acts like a duck we might as well call it a duck.

jsumners:
I don't use the `class` syntax. I do not agree with the decision to add it to the language. I would like to get rid of all the `this.` usage in this object as well, but for now I just want to get the code into a testable state and pulled out of one big pile of "client" code. It can be further improved later.

jsumners:

when opening a new issue.

Issue #51, opened 2012-01-17
Add connection timeout configuration for LDAP Client


thomasfr:
I want to authenticate to a LDAP Server and If it is not possbiel to reach the server within a given timeout, i want to fallback to another, local auth mechanism.
For this it is neccessary to have a timeout config option and maybe a timeout event fired when timeout reached.

PR #553, opened 2019-09-03
Added support for Virtual List View (vlv) control for browsing directory using paged search

soisik:
Hi,

Our app has to support large amount of LDAP entries in directory that our users can browse, search, and sort using some basic list and crud functionalities.

Since the current paging control will process all pages (one after another til it reach the end), it doesn't work for our use case: We need to be able to collect from LDAP directory just the one paginated list of LDAP objects that is requested, not all of the pages.

This PR contains support for the Virtual List View control that allow just this. Using this control alongside with the ServerSideSortingRequestControl, we are able to load only the requested page, which let us to build rich listing tools on top of large directory. It has been tested on OpenLDAP with the sssvlv overlay enabled.

Read https://tools.ietf.org/html/draft-ietf-ldapext-ldapv3-vlv-09 to understand how Virtual List View control works.

Some things to be aware of:

- This control only works if sent with a sss control (ServerSideSortingRequestControl). LDAP will reply error "#76 Control Error (Sort control is required with VLV)" if a search request a vlv control is performed with no sss control. This is specified in ietf.

- On OpenLDAP - I haven't check other directory - there is a limitation on how many searches with vlv control can be performed withing a single LDAP connection - see discussion here: https://www.openldap.org/lists/openldap-technical/201107/msg00006.html: After a few vlv requests using the same client, LDAP will reply with error "#51 BusyError: Other sort requests already in progress" until another client is used. So we have to make sure clients are not reused too many times for these requests by destroying/recreating them after a (few) vlv enabled search(es).

Some code example on how to use this control:

```
var ldapjs = require('ldapjs');

var opts = {
  filter: '(&(objectClass=person))',
  scope: 'sub',
  attributes: ['dn', 'sn', 'cn']
};

var controls = [];
controls.push(new ldapjs.ServerSideSortingRequestControl(
  { 
     value: {
	attributeType: 'uid',
	orderingRule: 'caseIgnoreOrderingMatch',
	reverseOrder: false
     }
   }
));
controls.push(new ldapjs.VirtualListViewRequestControl(
  { 
     value: {
	beforeCount:0,
	afterCount:2,
	targetOffset:1,
	contentCount:0
     }
   }
));

client.search('o=example', opts, function(err, res) {
  assert.ifError(err);

  res.on('searchEntry', function(entry) {
    console.log('entry: ' + JSON.stringify(entry.object));
  });
  res.on('searchReference', function(referral) {
    console.log('referral: ' + referral.uris.join());
  });
  res.on('error', function(err) {
    console.error('error: ' + err.message);
  });
  res.on('end', function(result) {
    for (let i = 0; i < result.controls.length; i++) {
      if (result.controls[i].type == ldapjs.VirtualListViewResponseControl.OID) {
        console.log('contentCount',result.controls[i]._value.contentCount);
        console.log('targetPosition',result.controls[i]._value.targetPosition);
      }
    }
  });
});
```


soisik:
Added: client test.

jsumners:
I hope to have time to study this PR this weekend, but it will likely be next weekend before I have the time.

jsumners:
@soisik I have been thinking that this PR brings up an interesting idea: module plugins. A small feature that allows registering a plugin which adds to the available LDAP codes and list of controls would make it simple to implement features like this one as separate modules. What do you think?

soisik:
@jsumners 
@jsumners 
I'll be happy to transform the VLV control into a plugin if such a system is implemented, but to be honest, I won't have time anytime soon to work on a plugin core system. Also, I'm quite new to node.js, and would not know what is the best way to do it.

jsumners:
I can probably work on this soon. I have to dig in and see how feasible it is, but I think it could be fairly trivial.

jsumners:
Well, with the way the code is currently structured my idea is not really feasible 😞. We'll just merge this in.

jsumners:

when opening a new issue.

PR #565, opened 2019-09-18
Fix escapedToHex

UziTech:
fixes [escapedToHex](https://github.com/ldapjs/node-ldapjs/blob/a113953e0d91211eb945d2a3952c84b7af6de41c/lib/filters/index.js#L167) when two escaped characters are next to each other

```js
escapedToHex("\\(\\)")
// before fix: "\\28\\)"
//  after fix: "\\28\\29"
```


jsumners:

when opening a new issue.

PR #569, opened 2019-09-29
Clean up options checking and add some tests

jsumners:
This was started as a branch to implement a plugin system for #553 but that failed. So this is just some updates around client options validation.


jsumners:

when opening a new issue.

PR #573, opened 2019-10-03
Fix test script

guimard:


jsumners:
This is covered by the `.taprc`:

https://github.com/ldapjs/node-ldapjs/blob/0ccbec6adba0d90bb485ee2869a18df6ca4d53d7/.taprc#L3-L4

It is done so because Windows doesn't pass the glob correctly to the suite runner for expansion. What error are you seeing that warrants this PR?

guimard:
OK, but here this file is not read here for an unknown reason. Not important, Debian build does not call `npm run test`

jsumners:

when opening a new issue.

Issue #340, opened 2016-01-20
client listener events documentations


iamtennislover:
Hi,
I would like to know what are the events that I can listen on for client? Is there any documentation for this?

```
 var client = ldap.createClient({
            url: testurl
        });
client.on('error', function(err) {console.log(err)})
```

dcolens:
I too would be interested to get some documentation about client events, there seems to be quite a lot:

- error
- destroy
- idle
- connectError
- setup
- resultError
- setupError
- socketTimeout
- end
- connect
- close
- timeout

while many are easy to guess, it would be great to get some documentation on the expected flows for these events :)

teemuniiranen:
Definitely some documentation is needed. I am currently struggling with close for example. I have no idea how to test and raise this event but without handling it is crashing my app.

```
at Client._onClose (/data/app/node_modules/ldapjs/lib/client/client.js:1272:19)
at Object.onceWrapper (events.js:293:19)
at emitOne (events.js:96:13)
at Socket.emit (events.js:191:7)
at TCP._handle.close [as _onclose] (net.js:511:12)
```

UziTech:
It might also be worth it to consolidate some of these events to make them more intuitive

jsumners:
👋

On February 22, 2023, we released [version 3](https://github.com/ldapjs/node-ldapjs/releases/tag/v3.0.0) of this library. As a result, we are closing this issue/pull request.

Please see [issue #839](https://github.com/ldapjs/node-ldapjs/issues/839) for more information, including how to proceed if you feel this closure is in error.

PR #574, opened 2019-10-11
Handle connection refused and connection timeouts gracefully

TPXP:
This pull request rebases the work from https://github.com/ldapjs/node-ldapjs/pull/433 to the next branch. It also adds tests to the two cases concerned by the PR:
- connection refused (LDAP server not started)
- connection timeout (firewall problem)

I took this as a opportunity to dive deeper into the code to understand what was actually causing the error to be thrown, I hope the code is a little cleaner than a blind handling of all errors (which caused other tests to fail). The issue was a poor handling of connection errors in the bind method. Since it's the first async method, it is the one that should report connection issues to the server, which it did not. Another issue was the handling of connection refused errors that caused a general error in ldapjs. Sending the exception to the bind callback avoids crashing the whole node process.

### DIFF
--- a/lib/client/client.js
+++ b/lib/client/client.js
@@ -120,8 +120,8 @@ function Client (options) {
 
   this.log = options.log.child({ clazz: 'Client' }, true)
 
-  this.on('error', function(err) {
-    self.cb(err);
+  this.on('error', function (err) {
+    self.cb(err)
   })
 
   this.timeout = parseInt((options.timeout || 0), 10)
@@ -176,7 +176,7 @@ module.exports = Client
  * Default handler for error callbacks when
  * one isn't set in the instance
  */
-Client.prototype.cb = function(err, ret) {
+Client.prototype.cb = function (err, ret) {
   if (err) {
     this.log.error('Caught exception:', err)
   } else {
@@ -299,7 +299,7 @@ Client.prototype.bind = function bind (name,
   })
 
   var self = this
-  this.cb = function(err, ret) {
+  this.cb = function (err, ret) {
     delete self.cb
     callback(err, ret)
   }
--- a/lib/client/client.js
+++ b/lib/client/client.js
@@ -282,13 +282,13 @@ Client.prototype.bind = function bind (name,
     controls: controls
   })
 
-  // While we are binding to the server, register the callback as error handler
+  // Connection errors will be reported to the bind callback too (useful when the LDAP server is not available)
   var self = this
   function callbackWrapper (err, ret) {
-    self.removeListener('error', callbackWrapper)
+    self.removeListener('connectError', callbackWrapper)
     callback(err, ret)
   }
-  this.addListener('error', callbackWrapper)
+  this.addListener('connectError', callbackWrapper)
 
   return this._send(req, [errors.LDAP_SUCCESS], null, callbackWrapper, _bypass)
 }
--- a/lib/client/client.js
+++ b/lib/client/client.js
@@ -120,6 +120,10 @@ function Client (options) {
 
   this.log = options.log.child({ clazz: 'Client' }, true)
 
+  this.on('error', function(err) {
+    self.cb(err);
+  })
+
   this.timeout = parseInt((options.timeout || 0), 10)
   this.connectTimeout = parseInt((options.connectTimeout || 0), 10)
   this.idleTimeout = parseInt((options.idleTimeout || 0), 10)
@@ -167,6 +171,18 @@ function Client (options) {
 }
 util.inherits(Client, EventEmitter)
 module.exports = Client
+
+/**
+ * Default handler for error callbacks when
+ * one isn't set in the instance
+ */
+Client.prototype.cb = function(err, ret) {
+  if (err) {
+    console.error('Caught exception:', err)
+  } else {
+    console.log('Unhandled output:', ret)
+  }
+}
 
 /**
  * Sends an abandon request to the LDAP server.
@@ -281,6 +297,12 @@ Client.prototype.bind = function bind (name,
     credentials: credentials || '',
     controls: controls
   })
+
+  var self = this
+  this.cb = function(err, ret) {
+    delete self.cb
+    callback(err, ret)
+  }
 
   return this._send(req, [errors.LDAP_SUCCESS], null, callback, _bypass)
 }
--- a/lib/client/client.js
+++ b/lib/client/client.js
@@ -120,10 +120,6 @@ function Client (options) {
 
   this.log = options.log.child({ clazz: 'Client' }, true)
 
-  this.on('error', function (err) {
-    self.cb(err)
-  })
-
   this.timeout = parseInt((options.timeout || 0), 10)
   this.connectTimeout = parseInt((options.connectTimeout || 0), 10)
   this.idleTimeout = parseInt((options.idleTimeout || 0), 10)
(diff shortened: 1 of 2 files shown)


lib/client/client.js
```diff
@@ -1005,6 +1013,8 @@ Client.prototype.connect = function connect () {
     // Communicate the last-encountered error
     if (err instanceof ConnectionError) {
       self.emit('connectTimeout', err)
+    } else if (err.code === 'ECONNREFUSED') {
+      self.emit('connectRefused', err)
```
UziTech: is this supposed to be `'connectError'` like above?

UziTech: All the tests pass when I comment out this line. Could you write a test that makes sure this emits the correct event?

UziTech:
> The issue was a poor handling of connection errors in the bind method. Since it's the first async method, it is the one that should report connection issues to the server

Technically `ldapjs.createClient` is an asynchronous function, it just uses an event emitter instead of a callback.

I'm curious why this wouldn't work to catch the error:

```js
const client = ldapjs.createClient(...)
client.on('error', (err) => {
	if (err.code === 'ECONNREFUSED') {
		// handle error
	}
})
client.bind(...)
```

UziTech:
maybe we should combine these errors and change it so `'ECONNREFUSED'` emits a `'connectError'`?

lib/client/client.js
```diff
@@ -1005,6 +1013,8 @@ Client.prototype.connect = function connect () {
...
     } else {
       self.emit('error', err)
```
UziTech: since this only emits errors in `connect` should this just be changed to emit 'connectError'?

FYI I believe this could be a breaking change.

TPXP: Those errors already emit a `connectError` event on https://github.com/ldapjs/node-ldapjs/blob/5afa6772099f11b36c92c677c73bd483ad2d7c80/lib/client/client.js#L803
This will emit another (more specific) error event.

On one hand, this can be a breaking change for codes handling connection errors with the `error` event. On the other side, the callback for `bind` will be called with the error so previously written code will still work.

TPXP:
@UziTech Thanks for the prompt feedback! <3

Granted, binding to the `error` event will work for handling connection refused errors, while binding to the `connectError` will handle connection timeouts. That's two events to bind to to make the code resilient and not **crash** (connection refused) or do nothing (timeout) if the server is not available (uncaught `error` events crash the node process, which is the motivation of #433). This requires advanced knowledge of the library events (undocumented), and I don't think users wanting a quick start want to go through all this for handling these basic events.

I don't find it particularly shocking to emit connection errors while logging into the server, for example that's what the [mysql](https://www.npmjs.com/package/mysql#establishing-connections) module will do.

UziTech:
> That's two events to bind to to make the code resilient and not crash

That is why I think we should make the connection refused error emit a `'connectError'` instead of an `'error'` so you will only need to handle one event.

There are also events for `'socketTimeout'`, `'setupError'`, `'connectTimeout'`, and `'timeout'`. Maybe we need to consolidate them to make them not so confusing. And probably document them better.

> for example that's what the [mysql](https://www.npmjs.com/package/mysql#establishing-connections) module will do.

I do think we should move the [`connect()`](https://github.com/ldapjs/node-ldapjs/blob/next/lib/client/client.js#L166) call out of the constructor so this might be less confusing.

the code would have to change to:

```js
const client = ldapjs.createClient(...)
client.connect((err) => {
	// handle connection errors
	client.bind(...)
})
```

jsumners:
Next release _will_ be a semver major. So some breaking changes are acceptable. I would like to keep them to a minimum, however. 

The `connect()` inside the constructor is quite annoying -- https://github.com/ldapjs/node-ldapjs/blob/5a31abaed0941e1bfd333917a9ab49d15d1712c3/test/client.test.js#L371-L387

But that might be a bit drastic of a change. A `connect` must be done before a `bind` can be performed. So it would introduce a complexity i bootstrapping a client. I can be convinced to make the change, but will need to be sufficiently documented.

Also, we *definitely* need documentation for the supported events -- https://github.com/ldapjs/node-ldapjs/issues/340

TPXP:
> That is why I think we should make the connection refused error emit a 'connectError' instead of an 'error' so you will only need to handle one event.
This current code already does this, it's just that it will emit another crash-causing `error` event.

Perhaps an acceptable compromise would be to move the connect method call in the bind if not yet called, so that it makes more sense to throw errors from there, but purists would still be able to catch connection errors by calling connect beforehand? This way, previously written code would still work

UziTech:
> Perhaps an acceptable compromise would be to move the connect method call in the bind if not yet called

Why not go further and include the connect and bind call in the search if not yet called?

```js
const client = ldapjs.createClient({username, password, ...});
client.search(...);
```

UziTech:
I'm thinking we should leave ldapjs as a low level library and not try to do too much.

What do you guys think about abolishing callback errors and just using event emitter for errors? I can't think of another library that uses event emitter and callbacks for errors.

It seems like the confusion comes from having the event emitter and callbacks.

Looking through the code I have found a few other places where it is confusing whether the error should be sent to the callback or the event emitter or both.

https://github.com/ldapjs/node-ldapjs/blob/ad451edc18d7768c3ddee1a1dd472d2bbafdae5e/lib/client/client.js#L933-L936

jsumners:
I think connection level events make sense being reported through the event emitter interface. Same with streaming parts of the API. But async methods that don’t deal with events should return callback style errors. 

I think most of this will shake out as we restructure the code into more maintainable chunks (files).

jsumners:
@TPXP do you have any thoughts on @UziTech's comments?

TPXP:
Hello there,

I agree that using callbacks and events for error handling is counter-intuitive, however I'd favor callback-style error handling for async methods as @jsumners suggested since it should make it easier to react to them in accordance with the context in which LDAPjs was called. For example, when building a web app, one would not want to avoid creating a client at every request and bind to events manually. If the callback gives an error object if something goes wrong, it would enable users to directly answer to this request

```js
const app = require('express')()

const ldap = ldapjs.createClient(...)
ldap.bind('dn=directory', ...)

app.get('/search', (req, res) => {
  ldap.search(req.body.query, ..., ..., (data, err) => {
    // I want any errors that occured during the search here (be it a bad request or a timeout, or whatever)
    if(err) res.end('oops, something went wrong')
    else res.json(data).end()
  }
}
```

Furthermore, if I were to bind to events in my request, I may catch events related to another request, which would be confusing. Or I would have to create a client at every request which would cause performance issues (lots of sockets).

TL;DR: Callback-style errors, please (which is actually sort of what this PR does for connection error events while bind-ing).

I don't think it's a good idea to call the `connect` method if it was not called previously for other methods than `bind`. Indeed, bind-ing is usually a compulsory step when working with a remote server so the fact that connect is only called here if needed is acceptable. It would "force" developers to confirm that they want to work anonymously since they would have to call `connect` manually. As said in [the docs](http://ldapjs.org/guide.html#bind), anonymous operations are usually _not_ what you want:
> So, lesson #1 about LDAP: unlike HTTP, it's connection-oriented; that means that you authenticate (in LDAP nomenclature this is called a bind), and all subsequent operations operate at the level of priviledge you established during a bind. You can bind any number of times on a single connection and change that identity. Technically, it's optional, and you can support anonymous operations from clients, but (1) you probably don't want that, and (2) most LDAP clients will initiate a bind anyway (OpenLDAP will), so let's add it in and get it out of our way.

So as long as it's said next to the bind method documentation that for connecting anonymously all you have to do is call connect instead of bind, it should be fine.

lib/client/client.js
jsumners: I think I am okay with the breaking change here since it will be going into a new major version. I have stated that the _intention_ is to not drastically alter the API, but some breaking changes are going to make it in.

jsumners:

when opening a new issue.

PR #584, opened 2019-12-03
refactor: resultError always emitted on errors

luddd3:
The check of self.listeners('resultError') is unnecessary because it will always result in an array which evaluates to true.


jsumners:
@luddd3 then wouldn't the more appropriate change be to check if that array's length is `> 0`?

jsumners:
Also, are you able to cover this change with a unit test?

lib/client/client.js
```diff
@@ -1150,7 +1150,7 @@ Client.prototype._sendSocket = function _sendSocket (message,
   var sentEmitter = false
 
   function sendResult (event, obj) {
-    if (event === 'error' && self.listeners('resultError')) {
+    if (event === 'error') {
```
UziTech:
```suggestion
    if (event === 'error' && self.listenerCount('resultError')) {
```

luddd3:
> Also, are you able to cover this change with a unit test?

This is already covered by a unit test. See `client.test.js` line 1438, "resultError handling".

lib/client/client.js
luddd3: Seems unnecessary. If there are no listeners, the emit is effectually a noop. It is not custom to look for listeners before emitting an event and it is not done anywhere else in this code base.

jsumners: Makes sense. I was only going on the intent implied by the original code in this change set.

jsumners:
@luddd3 can you do a force push an empty commit or something? GitHub Actions seems to be stuck and will not cancel (I have tried multiple times).

luddd3:
@jsumners I think something is wrong with the github action :(

jsumners:
I'll have to debug it this weekend :(

jsumners:
The CI issue was https://github.com/tapjs/node-tap/issues/620

Thank you for contributing. This has been released as `2.0.0-pre.4` and installable as `ldapjs@next`.

luddd3:
Thank you!

jsumners:

when opening a new issue.

PR #585, opened 2019-12-07
Update image reference

jsumners:


ThorvaldAagaard:
I have implemented your test in local code, and is testing against two different groups 

Just before the code

```
        const memberAttr = results[0].attributes.find(a => a.type === 'member')
        t.ok(memberAttr)
        t.ok(memberAttr.vals)
        t.type(memberAttr.vals, Array)
        t.is(memberAttr.vals.length, 2000)
```

I did add the following 

```
results[0].attributes.forEach(attr => {
                    console.log(attr.type)                    
                })
```
And the output from a small group is 

> objectClass
> cn
> description
> member
> distinguishedName
> instanceType
> whenCreated
> whenChanged
> uSNCreated
> uSNChanged
> name
> objectGUID
> objectSid
> sAMAccountName
> sAMAccountType
> groupType
> objectCategory
> dSCorePropagationData

and the output from the large group is 

> objectClass
> cn
> description
> member
> member;range=0-1499
> distinguishedName
> instanceType
> whenCreated
> whenChanged
> uSNCreated
> uSNChanged
> name
> objectGUID
> objectSid
> sAMAccountName
> sAMAccountType
> groupType
> objectCategory
> dSCorePropagationData

and for the large group the test failed as memberAttr.vals.length was 0

but using 
```
const memberAttr2 = entry.attributes.find(a => a.type === 'member;range=0-1499')
```
did return the first 1500 users in the group

jsumners:

when opening a new issue.

Issue #589, opened 2019-12-20
Error: Expired certificate causes untrappable error / breaks NodeJS.  (Fix included)


tfrancois:
Using NodeJS 10.13 and LDAPjs 2.0.0-pre.5, when connecting to an LDAP server with a TLS SSL certificate that has expired an error occurs that cannot be handled or trapped via catch error handling in NodeJS.  The fatal error that occurs is as follows:
```
Error: certificate has expired
    at TLSSocket.onConnectSecure (_tls_wrap.js:1055:34)
    at TLSSocket.emit (events.js:198:13)
    at TLSSocket.EventEmitter.emit (domain.js:448:20)
    at TLSSocket._finishInit (_tls_wrap.js:633:8)
Emitted 'error' event at:
    at Backoff.<anonymous> (...\node_modules\ldapjs\lib\client\client.js:1020:12)
    at Backoff.emit (events.js:198:13)
    at Backoff.EventEmitter.emit (domain.js:448:20)
    at Backoff.backoff (...\node_modules\backoff\lib\backoff.js:41:14)
    at ...\node_modules\ldapjs\lib\client\client.js:1002:15
    at f (...\node_modules\once\once.js:25:25)
    at TLSSocket.onResult (...\node_modules\ldapjs\lib\client\client.js:804:7)
    at Object.onceWrapper (events.js:286:20)
    at TLSSocket.emit (events.js:198:13)
    at TLSSocket.EventEmitter.emit (domain.js:448:20)
```
I was able to solve the issue by adding the following to **client.js** starting at lines 1006:
```
retry.on('fail', function (err) {
    if (self.destroyed) {
      // Silence any connect/setup errors if destroyed
      return
    }
    self.log.debug('failed to connect after %d attempts', failAfter)
    // Communicate the last-encountered error
    if (err instanceof ConnectionError) {
      self.emit('connectTimeout', err)
    } else if (err.code === 'ECONNREFUSED') {
      self.emit('connectRefused', err)
    } else if (err.code === 'CERT_HAS_EXPIRED' || err.code === 'UNABLE_TO_VERIFY_LEAF_SIGNATURE') {  // <-- THIS FIXED EXPIRED CERT ERROR
      self.emit('connectError', err)
    } else {
      self.emit('error', err)
    }
  })
```
The specific change that was added was:
```
[...]

PR #595, opened 2020-02-02
Handle expired TLS certificate errors in client

jonekdahl:
This PR contains the fix suggested by @tfrancois in #589. I am willing to add some tests as well, but I would need some guidance about how to set up such a test.

### DIFF
--- a/lib/client/client.js
+++ b/lib/client/client.js
@@ -1014,6 +1014,8 @@ Client.prototype.connect = function connect () {
       self.emit('connectTimeout', err)
     } else if (err.code === 'ECONNREFUSED') {
       self.emit('connectRefused', err)
+    } else if (err.code === 'CERT_HAS_EXPIRED' || err.code === 'UNABLE_TO_VERIFY_LEAF_SIGNATURE') {
+      self.emit('connectError', err)
     } else {
       self.emit('error', err)
     }


jsumners:
Thank you for submitting the PR. 

Let me think on how to add a test for it. I am thinking it would be easiest to just create an expired cert and start an ldapjs server instance with it. Then try to connect from a client.

lib/client/client.js
```diff
@@ -1014,6 +1014,8 @@ Client.prototype.connect = function connect () {
       self.emit('connectTimeout', err)
     } else if (err.code === 'ECONNREFUSED') {
       self.emit('connectRefused', err)
+    } else if (err.code === 'CERT_HAS_EXPIRED' || err.code === 'UNABLE_TO_VERIFY_LEAF_SIGNATURE') {
+      self.emit('connectError', err)
```
jsumners: @UziTech we already have this event at https://github.com/ldapjs/node-ldapjs/blob/8384a4736ff03bc6a09da5ee0e12d69eae02d277/lib/client/client.js#L793-L805

So this is not a brand new event.

UziTech: Yes but I don't understand why emitting those events is better than emitting an `error` event. `error` events are at least handled as errors if they are not handled by the user.

jonekdahl: I don't have an opinion on this. If you decide that connectError is appropriate, I will try to add some tests, otherwise I guess we can just close this PR. Either way is fine with me.

jsumners: @jonekdahl there's no reason to close the PR. Discussion is how code gets improved and accepted. 

@UziTech is referencing https://nodejs.org/dist/latest-v12.x/docs/api/events.html#events_error_events

It makes sense to fallback on that behavior because this library has a plethora of undocumented events. As a result, users are not listening for all of them correctly. Some try, but it's very difficult to do given the lack of documentation.

However, we can put a blocker on v2 that requires the documentation to be updated. I have added this PR to the project around improving documentation so that this event should not be forgotten in that work.

If we emit everything under the `error` event then the user has to do a lot of work to figure out what kind of error they have received and then determine how to handle it appropriately. At the moment, narrowing down errors by event type is the most user friendly way of doing this, I think. If we had an error system like Fastify(https://github.com/fastify/fastify/blob/7fa4bddca07aac18e0600cff40d97b95a96ab970/lib/errors.js) then I would be more inclined to agree that all errors should just emit via `error`.

What do you think @UziTech?

UziTech: I'm not saying emitting other error events is the wrong way to do it, but I'm trying to understand why it would be better. The undocumented error events seems to be a big issue for users.

I like the idea of creating our own errors like fastly and emitting them as an `error` event. If they are documented they should be just as easy to catch and they still crash the process if they aren't dealt with making them easier to debug. If someone doesn't subscribe to the `connectError` event it just fails silently right now.

UziTech: I feel like just documenting the error codes (e.g. `'ECONNREFUSED'`, `'CERT_HAS_EXPIRED'`, etc.) might be good enough. I don't think we need to go as far as creating our own errors.

jonekdahl: For version 2.0, perhaps the safest way to go is to keep whatever the 1.x behavior was, and document it. I would prefer to be on a stable (released) version of ldapjs and a supported Node LTS sooner rather than later, which is why I am trying to help out here.

If you agree, I can try to create a test to verify the 1.0.2 behavior on Node 8 and another test for the same behavior on the `next` branch on Node 10+.

jsumners: I agree with your assessment for the safest approach.

Please leave the `master` branch alone. Node 8 is EOL.

A test for this added to this PR would be fantastic. I think my initial idea around testing for it is the way to go. https://www.npmjs.com/package/node-forge could be helpful in this endeavor.

I truly appreciate you stepping in to help and get this library up to a new "stable" release.

UziTech:
I don't think this actually solves the issue in #589. That issue is about not being able to catch the error when the TLS certificate expires. But the OP at least knew what the error was being thrown because the error wasn't caught by subscribing to the `error` event.

With this PR the OP (or someone in a similar situation) wouldn't know what error was being thrown if they didn't know to subscribe to the `connectError` event because it will just fail silently.

If we do want to have error events other than `error` I think we should at least check for listeners on that event and throw an `error` event otherwise so it doesn't just fail silently.

Something like:

```js
} else if (err.code === 'CERT_HAS_EXPIRED' || err.code === 'UNABLE_TO_VERIFY_LEAF_SIGNATURE') {
	if (self.listenerCount('connectError') > 0) {
		self.emit('connectError', err)
	} else {
		// connectError is not caught
		self.emit('error', err)
	}
}
```

jsumners:
I think that's a fair compromise. In fact, I did the exact same thing in a work project just yesterday.

lib/client/client.js
jonekdahl: Do either of you happen to know what the 1.0.2 behavior is? If this is not a regression, fixing it should perhaps not be a blocker for a 2.0 release, right?

UziTech: The 1.0.2 behavior was to just emit an `error`. v2.0 release has many other breaking changes so I don't think it will be a problem to add one more.

jsumners: Regardless of whether or not v2 includes breaking changes, I'm am taking the stance of leaving v1 in its current known state and making v2 a sort of start fresh for the project.

jsumners: @jonekdahl can you make the change to check for listeners count on `connectError` and fallback to `error` if the count is `0`?

jonekdahl:
> I don't think this actually solves the issue in #589. That issue is about not being able to catch the error when the TLS certificate expires. But the OP at least knew what error was being thrown because the error wasn't caught by subscribing to the error event.

@UziTech This was what I assumed too at first, but it seems like you **can** catch the error by subscribing to the error event on the client:

```javascript
ldap = require('./lib/index');
const client = ldap.createClient({url: 'ldaps://expired.badssl.com:443'})

client.on('error', (err) => {
  console.log("error is caught");
});
```

When I run it (on the `next` branch):

```console
jon@node-ldapjs$ node client_with_listener.js
error is caught
jon@node-ldapjs$
```

lib/client/client.js
jonekdahl: @jsumners Sure, I could do that. Do you mean for connectError only, or for other custom error events like `connectTimeout`, `connectRefused` and `setupError` as well?

jsumners: Let's limit it to the error this PR resolves for now. A separate PR can cover all cases.

jonekdahl:
@UziTech @jsumners Given that it seems possible to listen to the `error` event (see [previous comment](https://github.com/ldapjs/node-ldapjs/pull/595#issuecomment-582270296)), I'm starting to question whether this change is needed.

jsumners:
Are you sure that is catching the expired cert error? I suspect it is catching a generic error due to trying to use LDAP to connect to an HTTP server.

UziTech:
That's what I was saying. The issue in #589 was not that the error wasn't catchable. It was that they didn't know how to catch it. The solution to the real problem should be more documentation around errors.

This code is able to catch the same error while still throwing the other errors.

```js
client.on('error', err => {
	if (err.code === 'CERT_HAS_EXPIRED' || err.code === 'UNABLE_TO_VERIFY_LEAF_SIGNATURE') {
		// handle cert expired
	} else {
		throw err;
	}
});
```

jonekdahl:
@jsumners Yes, I confirmed it by logging the error:

```javascript
ldap = require('ldapjs');
const client = ldap.createClient({url: 'ldaps://expired.badssl.com:443'})

client.on('error', (err) => {
  console.log("error is caught", err);
});
```

As you can see, it has the same `UNABLE_TO_VERIFY_LEAF_SIGNATURE` error code:

```console
$ node client_with_listener.js 
error is caught { Error: unable to verify the first certificate
    at TLSSocket.onConnectSecure (_tls_wrap.js:1055:34)
    at TLSSocket.emit (events.js:198:13)
    at TLSSocket._finishInit (_tls_wrap.js:633:8) code: 'UNABLE_TO_VERIFY_LEAF_SIGNATURE' }
```

@UziTech I agree with your last comment and that documentation would be helpful.

jsumners:
Okay, if we think documentation is the best path forward, feel free to close this PR and start a new one that addresses it via docs.

jsumners:

when opening a new issue.

PR #598, opened 2020-02-15
Add server.getConnections (fix #415)

### LINKED ISSUE #415: TypeError: server.getConnections is not a function
Hi all,

Thanks for this nice npm package. 
I want to access the number of concurrent connections on the server.
So, I used `server.connections`.

It works but returns the message :
```
(node:21622) DeprecationWarning: Server.connections property is deprecated. Use Server.getConnections method instead.
```
So I tried to use `server.getConnections()` but :
```
Unhandled Rejection : TypeError TypeError: server.getConnections is not a function
```

I don't know how to fix this, I looked up the code and getConneections is supposed to work !
Is it a bug ?

jonekdahl:
Add LDAP server method that delegates to node server.getConnection.

### DIFF
--- a/docs/server.md
+++ b/docs/server.md
@@ -39,9 +39,10 @@ Known compatible loggers are:
 Set this property to reject connections when the server's connection count gets
 high.
 
-### connections (getter only)
+### connections (getter only) - DEPRECATED
 
-The number of concurrent connections on the server.
+The number of concurrent connections on the server. This property is deprecated,
+please use server.getConnections() instead.
 
 ### url
 
@@ -95,6 +96,13 @@ Start a server listening for connections on the given file descriptor.
 This file descriptor must have already had the `bind(2)` and `listen(2)` system
 calls invoked on it. Additionally, it must be set non-blocking; try
 `fcntl(fd, F_SETFL, O_NONBLOCK)`.
+
+## Inspecting server state
+
+### server.getConnections(callback)
+
+The LDAP server API mirrors the [Node.js `server.getConnections` API](https://nodejs.org/dist/latest-v12.x/docs/api/net.html#net_server_getconnections_callback). Callback
+should take two arguments err and count.
 
 # Routes
 
--- a/lib/server.js
+++ b/lib/server.js
@@ -697,6 +697,10 @@ Server.prototype.address = function () {
   return this.server.address()
 }
 
+Server.prototype.getConnections = function (callback) {
+  return this.server.getConnections(callback)
+}
+
 Server.prototype._getRoute = function (_dn, backend) {
   assert.ok(dn)
 
--- a/test/server.test.js
+++ b/test/server.test.js
@@ -21,6 +21,30 @@ tap.test('basic create', function (t) {
   t.end()
 })
 
+tap.test('connection count', function (t) {
+  const server = ldap.createServer()
+  t.ok(server)
+  server.listen(0, 'localhost', function () {
+    t.ok(true, 'server listening on ' + server.url)
+
+    server.getConnections(function (err, count) {
+      t.error(err)
+      t.equal(count, 0)
+
+      const client = ldap.createClient({ url: server.url })
+      client.on('connect', function () {
+        t.ok(true, 'client connected')
+        server.getConnections(function (err, count) {
+          t.error(err)
+          t.equal(count, 1)
+          client.unbind()
+          server.close(() => t.end())
+        })
+      })
+    })
+  })
+})
+
 tap.test('properties', function (t) {
   const server = ldap.createServer()
   t.equal(server.name, 'LDAPServer')


jonekdahl:
CI fails on Windows, but it seems unrelated to these changes.

docs/server.md
```diff
@@ -96,6 +97,13 @@ This file descriptor must have already had the `bind(2)` and `listen(2)` system
 calls invoked on it. Additionally, it must be set non-blocking; try
 `fcntl(fd, F_SETFL, O_NONBLOCK)`.
 
+## Inspecting server state
+
+### server.getConnections(callback)
+
+The LDAP server API mirrors the node.js `server.getConnections` API. Callback
```
jsumners: Can you add a link to said documentation? LTS version should be fine.
```suggestion
The LDAP server API mirrors the [Node.js `server.getConnections` API](add link here). Callback
```

jonekdahl:
> Are you sure? It looks like the failure is in the connection count test added by this PR

@jsumners No, you're right. I just never scrolled that far down the build log, I saw this "error" near the top and assumed this was the issue:

```console
Transformation error for D:\a\node-ldapjs\node-ldapjs\lib\filters\substr_filter.js ; return original code
EPERM: operation not permitted, rename 'D:\a\node-ldapjs\node-ldapjs\node_modules\.cache\nyc\e5383ca64509b9fda11b23cc4c8c7e0c783716375fe22169c3543a879fd7bfdb.js.1073149460' -> 'D:\a\node-ldapjs\node-ldapjs\node_modules\.cache\nyc\e5383ca64509b9fda11b23cc4c8c7e0c783716375fe22169c3543a879fd7bfdb.js'
```

Any ideas on what the problem could be? I don't run Windows so I cannot easily debug it.

docs/server.md
jonekdahl: Sure. I linked to the `latest-v12.x` version.

jsumners:
> Any ideas on what the problem could be? I don't run Windows so I cannot easily debug it.

I also do not run Windows. I'll try to pull down GitHub's Docker image for it and run the tests when I get a chance. Could be a few days though.

UziTech:
I'm on windows. I'll try to check this out today

UziTech:
Just initial testing but it seems like using url instead of socket makes the test pass on windows.

```diff
 tap.test('connection count', function (t) {
   const server = ldap.createServer()
   t.ok(server)
-  server.listen(t.context.sock, function () {
+  server.listen(SERVER_PORT, function () {
     t.ok(true, 'server listening')
     server.getConnections(function (err, count) {
       t.error(err)
       t.equal(count, 0)

-      const client = ldap.createClient({ socketPath: t.context.sock })
+      const client = ldap.createClient({ url: 'ldap://127.0.0.1:' + SERVER_PORT })
       client.on('connect', function () {
         t.ok(true, 'client connected')
         server.getConnections(function (err, count) {
           t.error(err)
           t.equal(count, 1)
           client.unbind()
           server.close(() => t.end())
         })
       })
     })
   })
 })
```

jsumners:
That makes sense. Windows has many issues with Unix sockets.

jonekdahl:
Good find! I'll switch the test to use url instead of socket when I have some time over.

UziTech:
I have been trying to go down this rabbit hole but seem to be at a dead end. As far as I can tell, when connecting with a windows pipe this [`onconnection`](https://github.com/nodejs/node/blob/master/lib/net.js#L1530) function never gets called which is supposed to update the server connections. This seems to be an issue in ldapjs because if I just run `net.connect(pipe)` the `onconnection` method is called.

UziTech:
It seems the issue is that `socket.on('connect')` gets called before `onconnection` so the test finishes before the number of connections gets updated on the server.

adding a timout before the last `server.getConnections` call makes the test pass using `socketPath`

UziTech:
or you could wait for the server `connection` event.

```diff
 tap.test('connection count', function (t) {
   const server = ldap.createServer()
   t.ok(server)
   server.listen(t.context.sock, function () {
     t.ok(true, 'server listening')
     server.getConnections(function (err, count) {
       t.error(err)
       t.equal(count, 0)

       const client = ldap.createClient({ socketPath: t.context.sock })
-      client.on('connect', function () {
+      server.server.on('connection', function () {
         t.ok(true, 'client connected')
         server.getConnections(function (err, count) {
           t.error(err)
           t.equal(count, 1)
           client.unbind()
           server.close(() => t.end())
         })
       })
     })
   })
 })
```

jonekdahl:
Thanks for your investigation @UziTech, I went with the url variant and it seems to be working.

> This seems to be an issue in ldapjs because if I just run net.connect(pipe) the onconnection method is called.

Perhaps a new issue could be created for this problem?

jsumners:

when opening a new issue.

Issue #602, opened 2020-03-31
Search entries emitted too soon


ifroz:
When using the `client.search()` API, I encountered an issue, where my client sometimes missed the first few `searchEntry` events (or even the `end` event).
Apparently, if the LDAP server is much faster than the client, the client does not have enough time to bind event listeners in the callback.

In order to mitigate this issue, I published the fork [ldapjs-with-injectable-emitter](https://github.com/ifroz/node-ldapjs/commit/94716d1e46e86c620b1573a2e77f24cb75254c9e?fbclid=IwAR2eAFBC1x7NPZoJAjuL1HDMN38ruiZ-80EscW8XB-2XL1cwpJZPkcTuFJM),
which, in it's search options, accepts an optional `EventEmitter` instance, so that the user can set up all bindings before actually firing off the request. It can be used as follows:

```javascript
return new Promise((resolve, reject) => {
  const entries = []

  const emitter = new EventEmitter()
  emitter.on('searchEntry', entry => { users.push(entry) })
  emitter.on('error', err => reject(err))
  emitter.on('end', () => resolve(entries))

  client.search('ou=People', { emitter }, (err, res) => {
    if (err) reject(err)

    // Calling res.on(...) here might be too late
    res === emitter // => true
  })
})
```

This fixes my use-case, but the issue still remains for users of `v1.0.2`.
Therefore I'd like to open discussion regarding to either implement this approach in `ldapjs` or come up with another solution, so that `ldapjs` users can mitigate this caveat.

jsumners:
Is this an issue in `ldapjs@next`? Can you replicate it in the unit/integration tests?

ifroz:
- I experience the same issue using `v2.0.0-pre.5` (and `v1.0.2`)
- I have partial success reproducing the issue in unit tests, but [this `setTimeout`-based hanging test](https://github.com/ifroz/node-ldapjs/commit/5239f6c68827f2c25b4589089c199d15bb882412) aims to prove my point.
- In my use-case, I use node `v10.15.0` (I see that some tests on `master` hang in `node >= 10`)

jsumners:
New work should be developed against the `next` branch.

ifroz:
[...]

PR #610, opened 2020-05-13
Implement queueing search events until a listener appears

rkaw92:
This PR replaces the normal EventEmitter that's returned from `Client.search()` with a compatible "corked" version that will not start emitting events until a listener appears. It eliminates a race condition where some early events could be missed, preventing the user of the method from seeing the first few search results or, in extreme cases, the end boundary of the search.

For #602

### DIFF
--- a/test/client.test.js
+++ b/test/client.test.js
@@ -703,6 +703,25 @@ tap.test('search basic', function (t) {
   })
 })
 
+tap.test('GH-602 search basic with delayed event listener binding', function (t) {
+  t.context.client.search('cn=test, ' + SUFFIX, '(objectclass=*)', function (err, res) {
+    t.error(err)
+    setTimeout(() => {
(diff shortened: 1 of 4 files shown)


test/client.test.js
```diff
@@ -703,6 +703,25 @@ tap.test('search basic', function (t) {
   })
 })
 
+tap.test('GH-602 search basic with delayed event listener binding', function (t) {
+  t.context.client.search('cn=test, ' + SUFFIX, '(objectclass=*)', function (err, res) {
+    t.error(err)
+    setTimeout(() => {
```
jsumners, line 709: This could be flaky. But an alternative is not currently coming to me.

jsumners:
I intend to issue a release soon. Hopefully in the next few days.

jsumners:

when opening a new issue.

Issue #611, opened 2020-05-13
Searching with scope "sub" does not include baseDN on Samba 4.7.6


rkaw92:
Hi, we've found an issue (on current branch `next`) with subtree-scoped searches against Samba 4.7.6 LDAP not returning the base DN when they should.

Assuming `baseDN = 'CN=Users,DC=ad,DC=example,DC=org'` and a DIT with only the Users container (no objects underneath it):
* `ldapsearch -b "cn=Users,dc=ad,dc=example,dc=org" -s sub '(objectClass=container)'` :heavy_check_mark: works - returns exactly 1 entry (the Users container)
* `client.search(baseDN, { scope: 'base', filter: new EqualityFilter({ attribute: 'objectClass', value: 'container' }) })` :heavy_check_mark: works - returns exactly 1 entry
* `client.search(baseDN, { scope: 'sub', filter: new EqualityFilter({ attribute: 'objectClass', value: 'container' }) })` :x: does not work - returns exactly 0 entries - this should be impossible!

What is more concerning is that it seems that it behaves differently on Samba 4.11 (not sure why, but probably because of some opaque internal LDB handling changes).

We initially thought it was caused by #602, but even after fixing the race condition and using our fork of the library, the issue remained. The queries looked legit, but we decided to compare the generated BER on the wire with the actual messages sent by `ldapsearch`.

There's only 1 difference: **whitespace in DNs**:
* *ldapsearch* sends the DNs as-is, without processing: `CN=Users,DC=ad,DC=example,DC=org`
* `ldapjs` re-formats DNs like this: `cn=Users, dc=ad, dc=example, dc=org` - with spaces as separators!

What's strange is that RFC 4514 (*String Representation of Distinguished Names*) does not seem to imply spaces are valid RDN separators. In fact, the only separator it defines is `,` (comma, or U+002C).

Why it works with `scope: 'base'` is beyond me, but when we tested `ldapsearch` with a re-formatted baseDN with spaces inserted with scope `sub`, we got the same results as with `ldapjs` (0 results). This means the spaces really are extraneous.

This looks related to #151 and the defaults it established. I'm really surprised it hasn't caused a bigger number of issues till now.

[...]

PR #613, opened 2020-05-13
Remove spaces from DNs in SearchRequest

rkaw92:
This change fixes how DNs are serialized into BER for SearchRequest messages.
It drops spaces around commas, since it breaks some LDAP servers that expect
the DN to be separated by comma only as per RFC 4514. The old format of
spaced separators from LDAPv2 is thus no longer used.

For #611

### DIFF
--- a/lib/messages/search_request.js
+++ b/lib/messages/search_request.js
@@ -105,7 +105,13 @@ SearchRequest.prototype._parse = function (ber) {
 SearchRequest.prototype._toBer = function (ber) {
   assert.ok(ber)
 
-  ber.writeString(this.baseObject.toString())
+  // Format only with commas, since that is what RFC 4514 mandates.
+  // There's a gotcha here: even though it's called baseObject,
+  //  it can be a string or a DN object.
+  var formattedDN = dn.DN.isDN(this.baseObject)
+    ? this.baseObject.format({ skipSpace: true })
+    : this.baseObject.toString()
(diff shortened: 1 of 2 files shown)


lib/messages/search_request.js
```diff
@@ -105,7 +105,13 @@ SearchRequest.prototype._parse = function (ber) {
 SearchRequest.prototype._toBer = function (ber) {
   assert.ok(ber)
 
-  ber.writeString(this.baseObject.toString())
+  // Format only with commas, since that is what RFC 4514 mandates.
+  // There's a gotcha here: even though it's called baseObject,
+  //  it can be a string or a DN object.
+  var formattedDN = dn.DN.isDN(this.baseObject)
```
jsumners, line 111: I think we should fork `node-asn1` into the org after publishing `ldapjs@2` and potentially clean up ambiguous things like this.

jsumners:
@UziTech do you have any objections to this change?

lib/messages/search_request.js
rkaw92: @jsumners Does `node-asn1` actually have anything to do with this? It's not a bad idea in general, just missing the idea as it regards DN formatting.

jsumners: It's the module that provides BER encoding.

jsumners:

when opening a new issue.

Issue #612, opened 2020-05-13
Some tests fail if user has LDAP server running


rkaw92:
(This is a follow-up for the test failures I was having when developing #610)

If a server is running on localhost, some test assumptions are not met. In particular, this manages to connect simply because something is listening on the user's machine:
```js
tap.test('connection refused', function (t) {
  const client = ldap.createClient({
    url: 'ldap://0.0.0.0'
  })

  client.bind('cn=root', 'secret', function (err, res) {
    t.true(err)
    t.type(err, Error)
    t.equals(err.code, 'ECONNREFUSED')
    t.false(res)
    t.end()
  })
})
```

We should perhaps use a port that's less likely to be serving TCP. Given the project's audience, it's quite probable the developer is running an LDAP server on localhost.

Any port ideas? tcpmux - port 1? :sweat_smile:

jsumners:
I'd accept a PR that uses https://www.npmjs.com/package/get-port to fix this test.

jsumners:
⚠️ This issue has been locked due to age. If you have encountered a recent
problem that seems to be covered by this issue, please open a new issue.

Please include a [minimal reproducible example](https://en.wikipedia.org/wiki/Minimal_reproducible_example)
when opening a new issue.

PR #615, opened 2020-05-20
Use a random TCP port for testing timeouts/rejects

rkaw92:
This fixes a test failure when the host has something listening
on the LDAP TCP port. Previously, the test cases would assume
that the port was free and they would not connect.

For #612


jsumners:

when opening a new issue.

PR #630, opened 2020-07-16
Replace instanceof check with duck-typing in Filter#toBer mixin

### LINKED ISSUE #629: instanceof check fails in filter.js
Hi,
I'm running into a bug where:
* one package produces an LDAP connection and dependency-injects it into another's code
* the other package uses the connection to do `.search()` with some filters

The search throws an error:
```
TypeError: ber (BerWriter) required
```

Each package has its own copy of `ldapjs`:
```
xxx-our-software-that-creates-ldap-connection
├─┬ xxx-the-code-that-consumes-the-connection@1.0.0 (git+ssh://somewhere)
│ └─┬ ldapjs@2.0.0
│   └── asn1@0.2.4 
├─┬ ldapjs@2.0.0
│ └── asn1@0.2.4
```

This causes the [instanceof BerWriter check](https://github.com/ldapjs/node-ldapjs/blob/e54555ae0af74eae400e38eba5216c024e77945c/lib/filters/filter.js#L43) to fail - I suspect the `ber` instance is injected by the connection that executes the search (from the upper package), but the `instanceof` check from the mixin is done against the copy of `asn1` module that's owned by the lower package. Clearly, to JavaScript, these two are not the same prototypes as the prototype objects are created separately. Thus, `instanceof` returns false.

Broadly speaking, `instanceof` checks [are bound to cause this sort of problem](https://github.com/cujojs/when/issues/325) sooner or later when more than 1 version (or rather 1 copy) of a package is installed.

Is there concern that whatever we get as input for `toBer` in the filter mixin might be malformed / of a wrong type? If so, we need to duck-type this object to make sure it has the right methods, instead of relying on strict instanceof. Any objections?

rkaw92:
This fixes #629 where, if this package was used in 2 instances (e.g. depended upon by 2 different packages) where one passes a Client object to the other, they would not recognize each other's `BerWriter` instances and throw an error on `.search()` with filters.

### DIFF
--- a/lib/filters/filter.js
+++ b/lib/filters/filter.js
@@ -2,13 +2,9 @@
 
 // var assert = require('assert')
 
-var asn1 = require('asn1')
-
 var Protocol = require('../protocol')
 
 /// --- Globals
-
-var BerWriter = asn1.BerWriter
 
 var TYPES = {
   and: Protocol.FILTER_AND,
@@ -40,7 +36,7 @@ function isFilter (filter) {
 
 function mixin (target) {
   target.prototype.toBer = function toBer (ber) {
-    if (!ber || !(ber instanceof BerWriter)) { throw new TypeError('ber (BerWriter) required') }
+    if (!ber) { throw new TypeError('ber (BerWriter) required') }
 
     ber.startSequence(TYPES[this.type])
     ber = this._toBer(ber)
--- a/lib/filters/filter.js
+++ b/lib/filters/filter.js
@@ -45,7 +45,7 @@ function isBerWriter (ber) {
 
 function mixin (target) {
   target.prototype.toBer = function toBer (ber) {
-    if (!isBerWriter(ber)) { throw new TypeError('ber (BerWriter) required') }
+    if (isBerWriter(ber) === false) { throw new TypeError('ber (BerWriter) required') }
 
     ber.startSequence(TYPES[this.type])
     ber = this._toBer(ber)
--- a/lib/filters/filter.js
+++ b/lib/filters/filter.js
@@ -35,7 +35,8 @@ function isFilter (filter) {
 }
 
 function isBerWriter (ber) {
-  return (
+  return Boolean(
+    ber &&
     typeof (ber) === 'object' &&
     typeof (ber.startSequence) === 'function' &&
     typeof (ber.endSequence) === 'function'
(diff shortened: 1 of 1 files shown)


rkaw92:
Not sure what happened here with the CI:
> ##[error]Error: Command failed: git cat-file -p cdb4d834d7984e11bbc8687592b53c74fd133c7a

Random failure?

jsumners:
@rkaw92 can you rebase to latest master? I pushed #631 to fix the CI issue.

rkaw92:
@jsumners That helped, thanks!

lib/filters/filter.js
```diff
@@ -38,9 +34,17 @@ function isFilter (filter) {
...
+    typeof (ber) === 'object' &&
+    typeof (ber.startSequence) === 'function' &&
+    typeof (ber.endSequence) === 'function'
+  )
+}
+
 function mixin (target) {
   target.prototype.toBer = function toBer (ber) {
-    if (!ber || !(ber instanceof BerWriter)) { throw new TypeError('ber (BerWriter) required') }
+    if (!ber || !isBerWriter(ber)) { throw new TypeError('ber (BerWriter) required') }
```
jsumners, line 47:
```suggestion
    if (isBerWriter(ber) === false) { throw new TypeError('ber (BerWriter) required') }
```

And make the undefined check part `isBerWriter`.

UziTech: we should check for `null` in `isBerWriter` otherwise the error `Uncaught TypeError: Cannot read property 'startSequence' of null` will be thrown.

jsumners: Yeah, that's the coerced check for `!ber`. I just want to see it in the check function instead as another inlined condition.

rkaw92: Hey, I've just moved the falsiness check into the function to shorten the condition. Is this OK?

lib/filters/filter.js
```diff
@@ -44,7 +45,7 @@ function isBerWriter (ber) {
 
 function mixin (target) {
   target.prototype.toBer = function toBer (ber) {
-    if (!ber || !isBerWriter(ber)) { throw new TypeError('ber (BerWriter) required') }
+    if (!isBerWriter(ber)) { throw new TypeError('ber (BerWriter) required') }
```
jsumners, line 48: Nit: I prefer explicit checks over coercion negates where possible. I'm sure we have plenty of this shorthand syntax throughout the code, but I'd like to address them as we get to them.

```suggestion
    if (isBerWriter(ber) === false) { throw new TypeError('ber (BerWriter) required') }
```

jsumners:

when opening a new issue.

PR #634, opened 2020-07-21
add dependabot

UziTech:
Automate dependency updates with dependabot.

It would also be nice to automate releases with semantic-release.

We could also auto merge dev deps with dependabot.

### DIFF
--- a/.github/dependabot.yml
+++ b/.github/dependabot.yml
@@ -1,0 +1,10 @@
+version: 2
+updates:
+  - package-ecosystem: "github-actions"
+    directory: "/"
+    schedule:
+      interval: "daily"
+  - package-ecosystem: "npm"
+    directory: "/"
+    schedule:
+      interval: "daily"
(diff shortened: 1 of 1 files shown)


jsumners:
> It would also be nice to automate releases with semantic-release.

Not worth it in my opinion. There are so few releases that it would be more work to setup and maintain an automated publish system than to just do it by hand occasionally. Particularly if 2FA is required for publishing.

.github/dependabot.yml
```diff
@@ -0,0 +1,10 @@
+version: 2
+updates:
+  - package-ecosystem: "github-actions"
+    directory: "/"
+    schedule:
+      interval: "daily"
+  - package-ecosystem: "npm"
+    directory: "/"
+    schedule:
+      interval: "daily"
```
jsumners, line 10:
```suggestion
      interval: "weekly"
  - package-ecosystem: "npm"
    directory: "/"
    schedule:
      interval: "weekly"
```

I already get a _lot_ of notifications for package updates. I'd rather not have this project spamming as often as others.

jsumners:

when opening a new issue.

PR #642, opened 2020-07-22
Update dependabot.yml

jsumners:
+ Check for updates on a day when there is likely more time for them to be dealt with
+ Only apply version changes if they are not already covered by the configured qualifiers


jsumners:

when opening a new issue.

Issue #645, opened 2020-07-27
Spaces in searchResEntry objectName 


spookiej:
Hi, I found an issue (for me at least) where when using node-ldapjs as an LDAP server it was placing spaces in between the commas in the DN when replying to a request. The system I have it integrated did not like this as it treated them as malformed.

I did notice that there was another similar issue https://github.com/ldapjs/node-ldapjs/issues/611 reported and fixed, but on the searchRequest side.

The reason I am writing this issue, is to understand whether the fix I made for myself is a) in line with the LDAP standards, and b) something you would like PR'd into the library.

I copied the fix from https://github.com/ldapjs/node-ldapjs/pull/613, and ported it over to the `lib/messages/search_entry.js` library and it fixed my issue. 

Commit here: https://github.com/spookiej/node-ldapjs/commit/6efc5e0c43a388fef3c5f4d40ffd1d2805d1e4e7

The second issue I encountered, was the returned objectName being converted into lower case, even though the DN was camel case.

This wasn't so much an issue as my client had no problems with this, but I wanted to fix it so that it was correct behaviour for my client.

Commit here: https://github.com/spookiej/node-ldapjs/commit/fd119f592c273193783e0315687e7961a6f309e9

Would be great to get your feedback on these changes/issues.

jsumners:
Regarding the `, `: yes, please submit a PR to correct this issue. The spec is clear that a space should not follow a comma in the RDN path. This is a semver fix change IMHO.

Regarding case sensitivity: my reading of the specs indicates that case sensitivity is left up to the server implementation. I am fine with parroting back the attribute names as they were given. However, this will be a semver major change.

spookiej:
Hey @jsumners - thanks for getting back to me so quickly.

I've submitted a PR #646 for the comma space issue.

I'll look at raising a seperate PR for removing the lowercase transformation on the response.

jsumners:
Released as v2.1.1

jsumners:
⚠️ This issue has been locked due to age. If you have encountered a recent
problem that seems to be covered by this issue, please open a new issue.

Please include a [minimal reproducible example](https://en.wikipedia.org/wiki/Minimal_reproducible_example)
when opening a new issue.

PR #646, opened 2020-07-27
Remove spaces in searchResEntry objectName

spookiej:
This PR addresses issue #645

This change fixes how DNs are serialised into BER for searchResEntry messages, and is related to the following other issues and PRs (#611, #613).

The change is required as it breaks clients that are expecting the reply to be comma separated only and not comma space `, `


jsumners:

when opening a new issue.

PR #647, opened 2020-07-28
Remove lowercase conversion on searchResEntry objectName

stenh0use:
This PR addresses issue #645

This change fixes how DNs are formatted for searchResEntry messages.

Existing behavior forces the object class in the response to be in lower case regardless of how it is stored in the server.

eg. data in the database that is stored like the following:
```
camelCase=camelCase,cn=foo,o=test
```
This will be returned to the requester in the following format:
```
camelcase=camelCase,cn=foo,o=test
```

The new behavior once this PR is merged will allow for the server to send the response in the same format that the data is stored in the database.

eg.
```
camelCase=camelCase,cn=foo,o=test
```

### DIFF
--- a/lib/dn.js
+++ b/lib/dn.js
@@ -46,14 +46,13 @@ RDN.prototype.set = function rdnSet (name, value, opts) {
   assert.string(value, 'value (string) required')
 
   var self = this
-  var lname = name.toLowerCase()
-  this.attrs[lname] = {
+  this.attrs[name] = {
     value: value,
     name: name
   }
   if (opts && typeof (opts) === 'object') {
     Object.keys(opts).forEach(function (k) {
-      if (k !== 'value') { self.attrs[lname][k] = opts[k] }
+      if (k !== 'value') { self.attrs[name][k] = opts[k] }
     })
   }
 }


spookiej:
I'll fix the build once https://github.com/ldapjs/node-ldapjs/pull/646 is merged

spookiej:
@jsumners what do you mean by non-breaking manner? 

This change means that any DN that is not purely lower case will be returned by the server to the requester as they sent it.

We could potentially introduce a configurable option to default to existing behaviour but enable this to be turned on. I am not very familiar with the core code base, are there any examples of configurable features you could point me to? I can look at implementing something that does not change existing behaviour.

jsumners:
> This change means that any DN that is not purely lower case will be returned by the server to the requester as they sent it.

Current implementation would expect the lowercasing to happen. Changing this behavior is a breaking change.

> We could potentially introduce a configurable option to default to existing behaviour but enable this to be turned on. I am not very familiar with the core code base, are there any examples of configurable features you could point me to? I can look at implementing something that does not change existing behaviour.

Your familiarity is probably on par with mine. I am indeed suggesting an option that can be passed along to the RDN constructor that would enable case parroting.

spookiej:
No worries, I'll look into it when I get the chance.

jwilleke:
This is a Bad Idea.
LDAP, which is a protocol, SHOULD NOT be manipulating values stored within the LDAP server.

DNs are the construction of RDNs that make up the DN and are LDAP Attribute Values and should NOT be manipulated by the Access Protocol.
This would imply that a results from a user store would be different based on access protocol. (ie. SQL vs LDAP)

spookiej:
@jwilleke I probably didn't explain my PR very well.

What you are saying is true, and is actually the problem that this PR addresses. The data in the database is being converted to lowercase regardless of how it is stored.

calvincheng8:
Yes this change makes sense. 

Agree with @jsumners that the current behavior should be preserved, users should be given an option to use this feature.

@spookiej Please rephrase the PR description, I did not understand what you meant until I read the actual code changes and the  discussions.

jsumners:
👋

Issue #326, opened 2015-12-07
How to use client.starttls()


lukelafountaine:
I am having a little trouble figuring out how to use the client starttls() function. I am creating the client this way: 

```
client = ldap.createClient({
    url: config.ldap.url
});
```

The client is created correctly and upon calling the starttls function as 

```
client.starttls({}, function(err) {
    console.log('inside the starttls's callback');
});
```

However, the callback function never gets called. I get the error : `[TypeError: Object.keys called on non-object]`. In the Node.js TLS documentation it mentions that on connecting, if the certificate authority is left blank it will default to the system certificates, and I was assuming that this would act in the same way. If there were any thoughts on what is going wrong here, that would be great.

pfmooney:
Please post the full stacktrace from the error.

aneelaSaleem:
Please look at issue #307 
It might help you solve the problem

aneelaSaleem:
Please look at issue #307
It might help you solve the problem

pfmooney:
@aneelaSaleem It does not appear that this issue has anything to do with the difficulties you experienced in #307.  Please do not post references to tickets unless you know them to be related.

lukelafountaine:
TypeError: Object.keys called on non-object
      at Function.keys (native)
      at wrapper (/usr/src/app/node_modules/ldapjs/node_modules/once/node_modules/wrappy/wrappy.js:27:14)
      at Client.starttls (/usr/src/app/node_modules/ldapjs/lib/client/client.js:880:14)
      at AuthenticationHelper.validateToken (/usr/src/app/helpers/authentication.js:97:12)
      at Server.AuthenticationCheck.checkAuth (/usr/src/app/middlewares/authenticationCheck.js:25:14)
      at next (/usr/src/app/node_modules/restify/lib/server.js:906:30)
      at f (/usr/src/app/node_modules/restify/node_modules/once/once.js:17:25)
      at b (domain.js:183:18)
      at Domain.run (domain.js:123:23)
      at Server._run (/usr/src/app/node_modules/restify/lib/server.js:959:7)
      at Server._handle (/usr/src/app/node_modules/restify/lib/server.js:719:14)
      at Server.onRequest (/usr/src/app/node_modules/restify/lib/server.js:326:14)
      at Server.emit (events.js:98:17)
      at HTTPParser.parser.onIncoming (http.js:2109:12)
[...]

PR #650, opened 2020-08-07
added controls variable in starttls example

undoingtech:
Regarding issue #326.  Added controls variable and parameter to the starttls example.

### DIFF
--- a/docs/client.md
+++ b/docs/client.md
@@ -386,7 +386,9 @@ Example:
       ca: [fs.readFileSync('mycacert.pem')]
     };
 
-    client.starttls(opts, function(err, res) {
+    var controls = client.controls;
+
+    client.starttls(opts, controls, function(err, res) {
       assert.ifError(err);
 
       // Client communication now TLS protected


docs/client.md
```diff
@@ -386,7 +386,9 @@ Example:
       ca: [fs.readFileSync('mycacert.pem')]
     };
 
-    client.starttls(opts, function(err, res) {
+    var controls = client.controls;
```
jsumners, line 389: Can we add a note about why we are doing this?

undoingtech:
I think I made a mistake with this PR.  I was under the impression that the `controls` parameter was required.  Technically the `controls` parameter is required as currently implimented, but  it should be optional (as according to the "Common patterns" section in client API documentation).  

I would rather fix the code than update the documentation.

I suggest this PR be closed/rejected.  I will start a new PR where I fix the starttls function so the controls parameter is optional.

jsumners:

when opening a new issue.

Issue #408, opened 2016-12-07
Support multiple servers in the client


silverwind:
`url` could take a array of URLs and LDAP operations could then be done in parallel, with the fastest one winning. Maybe a option to do sequential connection attempts might also make sense.

silverwind:
Made a [wrapper](https://github.com/silverwind/multi-ldap) that supports multiple servers. I had to make `createClient` async, which makes it incompatible to merge, but maybe you have a better idea on how to go about it.

gurisko:
Is there any reason why ldapjs wouldn't want to support multiple server URLs?

davama:
this would be awesome!

jsumners:
⚠️ This issue has been locked due to age. If you have encountered a recent
problem that seems to be covered by this issue, please open a new issue.

Please include a [minimal reproducible example](https://en.wikipedia.org/wiki/Minimal_reproducible_example)
when opening a new issue.

PR #658, opened 2020-09-16
Client support multiple servers

wision:
Resolves: https://github.com/ldapjs/node-ldapjs/issues/408

### DIFF
--- a/docs/client.md
+++ b/docs/client.md
@@ -14,7 +14,7 @@ The code to create a new client looks like:
 
     var ldap = require('ldapjs');
     var client = ldap.createClient({
-      url: 'ldap://127.0.0.1:1389,127.0.0.2:1389'
+      url: ['ldap://127.0.0.1:1389', 'ldap://127.0.0.2:1389']
     });
 
 You can use `ldap://` or `ldaps://`; the latter would connect over SSL (note
--- a/docs/client.md
+++ b/docs/client.md
@@ -14,7 +14,7 @@ The code to create a new client looks like:
 
     var ldap = require('ldapjs');
     var client = ldap.createClient({
-      url: 'ldap://127.0.0.1:1389'
+      url: 'ldap://127.0.0.1:1389,127.0.0.2:1389'
     });
 
 You can use `ldap://` or `ldaps://`; the latter would connect over SSL (note
@@ -24,7 +24,7 @@ client is:
 
 |Attribute      |Description                                                |
 |---------------|-----------------------------------------------------------|
-|url            |A valid LDAP URL (proto/host/port only)                    |
+|url            |A valid LDAP URL(s) (proto/host(s)/port(s)                 |
 |socketPath     |Socket path if using AF\_UNIX sockets                      |
 |log            |A compatible logger instance (Default: no-op logger)       |
 |timeout        |Milliseconds client should let operations live for before timing out (Default: Infinity)|
--- a/docs/client.md
+++ b/docs/client.md
@@ -24,7 +24,7 @@ client is:
 
 |Attribute      |Description                                                |
 |---------------|-----------------------------------------------------------|
-|url            |A valid LDAP URL(s) (proto/host(s)/port(s)                 |
+|url            |A string or array of valid LDAP URL(s) (proto/host/port)   |
 |socketPath     |Socket path if using AF\_UNIX sockets                      |
 |log            |A compatible logger instance (Default: no-op logger)       |
 |timeout        |Milliseconds client should let operations live for before timing out (Default: Infinity)|
--- a/docs/client.md
+++ b/docs/client.md
@@ -34,6 +34,13 @@ client is:
 |strictDN       |Force strict DN parsing for client methods (Default is true)|
 |reconnect      |Try to reconnect when the connection gets lost (Default is false)|
 
+### url
+This parameter takes a single connection string or an array of connection strings
+as an input. In case an array is provided, the client tries to connect to the
+servers in given order. To achieve random server strategy (e.g. to distribute
+the load among the servers), please shuffle the array before passing it as an
+argument.
+
 ### Note On Logger
 
 A passed in logger is expected to conform to the [Bunyan](https://www.npmjs.com/package/bunyan)
--- a/lib/client/client.js
+++ b/lib/client/client.js
@@ -793,8 +793,8 @@ Client.prototype.connect = function connect () {
 
   // Establish basic socket connection
   function connectSocket (cb) {
-    self._nextServer = (self._nextServer + 1) % self.urls.length
     var server = self.urls[self._nextServer]
+    self._nextServer = (self._nextServer + 1) % self.urls.length
 
     cb = once(cb)
 
--- a/lib/client/client.js
+++ b/lib/client/client.js
@@ -798,9 +798,9 @@ Client.prototype.connect = function connect () {
     self.url = self.urls[self._nextServer]
 
     if (self.url) {
-      self.host = url.hostname
-      self.port = url.port
-      self.secure = url.secure
+      self.host = self.url.hostname
+      self.port = self.url.port
+      self.secure = self.url.secure
     }
 
     cb = once(cb)
@@ -831,13 +831,12 @@ Client.prototype.connect = function connect () {
       setupClient(cb)
     }
 
-    var port = (self.urls.length && self.urls[self._nextServer].port) || self.socketPath
-    var hostname = (self.urls.length && self.urls[self._nextServer].hostname) || undefined
+    var port = (self.port || self.socketPath)
     if (self.secure) {
-      socket = tls.connect(port, hostname, self.tlsOptions)
+      socket = tls.connect(port, self.host, self.tlsOptions)
       socket.once('secureConnect', onConnect)
     } else {
-      socket = net.connect(port, hostname)
+      socket = net.connect(port, self.host)
       socket.once('connect', onConnect)
     }
     socket.once('error', onResult)
--- a/lib/client/client.js
+++ b/lib/client/client.js
@@ -111,8 +111,7 @@ function Client (options) {
 
   var self = this
   this.urls = options.url ? [].concat(options.url).map(url.parse) : []
-  // select random server at the start
-  this._nextServer = Math.floor(Math.random() * this.urls.length)
+  this._nextServer = 0
   // updated in connectSocket() after each connect
   this.host = undefined
   this.port = undefined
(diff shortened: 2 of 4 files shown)


docs/client.md
```diff
@@ -14,7 +14,7 @@ The code to create a new client looks like:
 
     var ldap = require('ldapjs');
     var client = ldap.createClient({
-      url: 'ldap://127.0.0.1:1389'
+      url: 'ldap://127.0.0.1:1389,127.0.0.2:1389'
```
UziTech, line 17: IF we are going to support multiple servers, the url parameter should be an array.

UziTech (changes requested):

lib/client/client.js
```diff
@@ -792,6 +797,8 @@ Client.prototype.connect = function connect () {
 
   // Establish basic socket connection
   function connectSocket (cb) {
+    self.nextServer = (self.nextServer + 1) % self.servers.length
+    self.url = self.servers[self.nextServer]
```
UziTech, line 801: This seems like a bad idea to use a mutating parameter when dealing with async code.

What stops this from changing before one of the async functions uses it?

docs/client.md
wision: I can add support for arrays, but having single string as a connection string with multiple servers is quite usual.. some examples:
mongodb: https://github.com/mongodb/node-mongodb-native/blob/master/src/connection_string.ts#L139-L149
zookeeper: https://github.com/alexguan/node-zookeeper-client/blob/master/lib/ConnectionStringParser.js#L51-L55

jsumners: Because they do it doesn't mean it is a good idea. I would much prefer an array of URLs.

lib/client/client.js
```diff
@@ -110,12 +110,17 @@ function Client (options) {
...
-  this.secure = _url ? _url.secure : false
-  this.url = _url
+  this.servers = []
+  if (options.url) {
+    var [, protocol, hosts, path] = options.url.match(/^([a-zA-Z]+:\/\/)([^/]*)(.*)/)
+
+    hosts.split(',').forEach((host) => {
+      this.servers.push(url.parse(`${protocol}${host}${path}`))
+    })
+  }
```
jsumners, line 120: This code can be simplified to `this.servers = Array.isArray(opts.url) ? opts.url : [opts.url]` when we do not accept comma separated strings.

UziTech:
@wision what is the purpose of the multiple servers in this PR. Is it for redundancy, lower latency, other?

jsumners:
I think this PR is looking pretty good. We need to address feedback and add some tests.

wision:
> @wision what is the purpose of the multiple servers in this PR. Is it for redundancy, lower latency, other?

The reason is redundancy / failover. We have three ldap servers and it became important that we keep serving requests even if one servers is down (e.g. maintenance or similar).

lib/client/client.js
wision: Both (`self.nextServer` & `self.url`) are updated only at the start of `connectSocket` function. Reading both is done in synchronous way (https://github.com/wision/node-ldapjs/blob/multi-servers/lib/client/client.js#L838-L839 and https://github.com/wision/node-ldapjs/blob/multi-servers/lib/client/client.js#L865, where `initSocket` is called synchronously in `connectSocket` https://github.com/wision/node-ldapjs/blob/multi-servers/lib/client/client.js#L848)

wision:
Anyone has an idea how to fix that windows test? Before adding `connectTimeout` it happened that sometimes the test passed and sometimes it failed (I'm using debian bullseye & nodejs 14). After adding `connectTimeout` I ran the test ~20 times and it always passed.. but it seems that windows still has some problems.

UziTech:
> The reason is redundancy / failover.

Ok. When I first saw "multiple servers" I thought it would be for lower latency. I thought it would try to connect to all of the servers at once and choose the first one that responds.

lib/client/client.js
UziTech: It looks like `self.url` is only used in [`initSocket`](https://github.com/wision/node-ldapjs/blob/multi-servers/lib/client/client.js#L863) there is no reason for it to be global especially since it is not necessarily the url that will be used to connect.

If someone calls `client.connect()` then `client.url` could be the wrong url.

`self.nextServer` looks like it does have to be global but it doesn't have to be public.

UziTech (changes requested):

lib/client/client.js
```diff
@@ -110,12 +110,23 @@ function Client (options) {
   EventEmitter.call(this, options)
 
   var self = this
-  var _url
-  if (options.url) { _url = url.parse(options.url) }
-  this.host = _url ? _url.hostname : undefined
-  this.port = _url ? _url.port : false
-  this.secure = _url ? _url.secure : false
-  this.url = _url
+  this.servers = []
```
UziTech, line 113: Just nitpicking but would `this.urls` be more appropriate? since it is a public property it would be better to get it right now so we don't need a major bump to change it later.

lib/client/client.js
```diff
@@ -829,7 +840,7 @@ Client.prototype.connect = function connect () {
       socket.once('connect', onConnect)
     }
     socket.once('error', onResult)
-    initSocket()
+    initSocket(self.url)
```
UziTech, line 843: This should be `initSocket(self.urls[self._nextServer])` and `self.url` should be `undefined` until the server connection is complete and we know which url we are actually connected to. Probably could be done in the `connectSocket` [callback](https://github.com/wision/node-ldapjs/blob/multi-servers/lib/client/client.js#L1009)

UziTech (changes requested):

lib/client/client.js
```diff
@@ -820,16 +825,17 @@ Client.prototype.connect = function connect () {
       setupClient(cb)
     }
 
-    var port = (self.port || self.socketPath)
+    var port = (server && server.port) || self.socketPath
+    var host = server && server.hostname
     if (self.secure) {
```
UziTech, line 830: This will always be `false`

UziTech (changes requested):

lib/client/client.js
```diff
@@ -975,6 +988,9 @@ Client.prototype.connect = function connect () {
       maxDelay: this.reconnect.maxDelay
     })
     failAfter = this.reconnect.failAfter
+    if (this.urls.length > 1 && failAfter) {
+      failAfter *= this.urls.length
+    }
```
UziTech, line 993: if `reconnect` option is not specified `failAfter` will be equal to `1` so only the first url will be tried.

UziTech (changes requested):

lib/client/client.js
```diff
@@ -110,12 +110,14 @@ function Client (options) {
...
-  this.port = _url ? _url.port : false
-  this.secure = _url ? _url.secure : false
-  this.url = _url
+  this.urls = options.url ? [].concat(options.url).map(url.parse) : []
+  // select random server at the start
+  this._nextServer = Math.floor(Math.random() * this.urls.length)
+  // updated in connectSocket() after each connect
+  this.host = undefined
+  this.port = false
+  this.secure = false
```
UziTech, line 119: another nit but should these be `undefined` until they are defined?

lib/client/client.js
wision: Good point.. I'd say that while using multiple servers you always want to reconnect (meaning that disabling `reconnect` option does not make sense), but guess I shouldn't modify this behavior and just update `failAfter` to the number of servers in use.

test/client.test.js
```diff
@@ -379,6 +378,20 @@ tap.test('createClient', t => {
     }
   })
 
+  t.test('url array is correctly assigned', async t => {
+    getPort().then(function (unusedPortNumber) {
+      const client = ldap.createClient({
+        url: [
+          `ldap://0.0.0.0:${unusedPortNumber}`,
+          `ldap://0.0.0.1:${unusedPortNumber}`
```
UziTech, line 386: It looks like the tests are failing because the server can't connect to 0.0.0.0
```suggestion
          `ldap://127.0.0.1:${unusedPortNumber}`,
          `ldap://127.0.0.2:${unusedPortNumber}`
```

UziTech (changes requested):

test/client.test.js
wision: Weird that sometimes it succeeds sometimes it fails. I checked for active handles after the test is over and sometimes the sockets stay in paused mode (hence the fail, because the test didn't terminate itself).

UziTech (approved):
LGTM

lib/client/client.js
```diff
@@ -110,12 +110,14 @@ function Client (options) {
...
   var self = this
-  var _url
-  if (options.url) { _url = url.parse(options.url) }
-  this.host = _url ? _url.hostname : undefined
-  this.port = _url ? _url.port : false
-  this.secure = _url ? _url.secure : false
-  this.url = _url
+  this.urls = options.url ? [].concat(options.url).map(url.parse) : []
+  // select random server at the start
+  this._nextServer = Math.floor(Math.random() * this.urls.length)
```
jsumners, line 115: Should it be random? Or should we try to connect to each server sequentially? I can easily envision someone supplying an array of servers ordered in their preferred connect order.

UziTech: So for multiple server strategies we could have:

1. Redundancy random order
2. Redundancy prefered order
3. Lowest latency

Maybe we need a way for people to instruct ldapjs which one they want?

jsumners: That's a lot to ask for this PR and the codebase as it currently exists. 2 and 3 are easily met by the user determining viability before constructing the array to pass in (i.e. they ping the things and create the array in latency order).

If you two think random is the way to go, then we'll ship it as-is. Otherwise, the only change I'd make before merging is using the array sequentially as, in my opinion, it is going to be the case assumed by most users.

wision: Should this configuration be a part of `reconnect` option? Or should I create new option `connectStrategy`?

UziTech: I agree the `url` array could be randomized each time it is sent to ldapjs if someone wants it to be randomized.

We probably don't need other strategies, just the preferred order.

jsumners (approved):
LGTM

I'll merge release likely this weekend.

UziTech:
It might be worth it to create a demo in the docs of how someone could randomize the array before sending it to ldapjs in order to achieve the other strategy.

jsumners:

when opening a new issue.

PR #674, opened 2020-10-31
update devdependency standard to v16

UziTech:
Update standard and fix linting issues.

Alternatively (if for some reason we want to use `var`) we could switch the linter to eslint and use eslint-config-standard and turn some rules off.


jsumners (approved):
I find it very difficult to be confident that there are not any scoping issues here. But the tests are passing, so 🤷‍♂️

UziTech:
Ya I think the tests or other linting rules would have caught any issues. There were a few places I had to manually fix because `--fix` wouldn't fix them automatically.

jsumners:

when opening a new issue.

Issue #676, opened 2020-11-11
Incomplete compatibility with Pino logger


telepsis:
There are conditional tests using logging methods that behave differently when using Pino vs Bunyan, e.g. like this one in `lib/server.js`:

```if (log.trace()) { log.trace('data on %s: %s', c.ldap.id, util.inspect(data)) }```

The Pino methods always return void while Bunyan's eval to true. Further, the arg-less Pino method calls fire off a bare event to the output stream so we see a slew of message-less events.

It would seem that those tests are not necessary given the abstract-logger noop methods and the log level options built into the other packages.

Node source to reproduce:
```
const log = require('pino')({level: 'trace'});
require('ldapjs').createServer({log: log}).listen(1389, 'localhost', () => log.info('Hello'));
```
Run it then call: `ldapsearch -H ldap://localhost:1389 -x -b o=my gawd=*`

Console output from nodejs is:
```
{"level":30,"time":1605124530404,"pid":450,"hostname":"LAPTOP-fed","msg":"Hello"}
{"level":10,"time":1605124534421,"pid":450,"hostname":"LAPTOP-fed","msg":"new connection from 127.0.0.1:55384"}
{"level":10,"time":1605124534423,"pid":450,"hostname":"LAPTOP-fed"}
{"level":10,"time":1605124534425,"pid":450,"hostname":"LAPTOP-fed"}
{"level":10,"time":1605124534425,"pid":450,"hostname":"LAPTOP-fed"}
{"level":20,"time":1605124534426,"pid":450,"hostname":"LAPTOP-fed"}
{"level":20,"time":1605124534431,"pid":450,"hostname":"LAPTOP-fed"}
{"level":10,"time":1605124534433,"pid":450,"hostname":"LAPTOP-fed"}
{"level":10,"time":1605124534433,"pid":450,"hostname":"LAPTOP-fed"}
{"level":10,"time":1605124534434,"pid":450,"hostname":"LAPTOP-fed"}
{"level":20,"time":1605124534434,"pid":450,"hostname":"LAPTOP-fed"}
{"level":20,"time":1605124534436,"pid":450,"hostname":"LAPTOP-fed"}
{"level":10,"time":1605124534437,"pid":450,"hostname":"LAPTOP-fed"}
{"level":10,"time":1605124534437,"pid":450,"hostname":"LAPTOP-fed"}
{"level":10,"time":1605124534437,"pid":450,"hostname":"LAPTOP-fed"}
[...]

PR #677, opened 2020-11-12
Remove use of logging methods in conditional tests

telepsis:
To address issue #676 .

Use of log methods (e.g. `log.trace()`) in conditional tests prior to actually logging messages yields undesirable behavior when using Pino as an alternative to Bunyan (and potentially other Bunyan-like packages).

1. Pino methods have void return values and always fail the conditional test
2. Pino method calls with no parameters still produce a logging event with several default values but no message field

Additionally, the default behavior for Bunyan and similar loggers is to filter events based on instance config options. The conditionals are not necessary.

### DIFF
--- a/lib/client/client.js
+++ b/lib/client/client.js
@@ -716,12 +716,12 @@ Client.prototype.starttls = function starttls (options,
          */
         secure.removeAllListeners('error')
         secure.on('data', function onData (data) {
-          if (self.log.trace()) { self.log.trace('data event: %s', util.inspect(data)) }
+          self.log.trace('data event: %s', util.inspect(data))
 
           self._tracker.parser.write(data)
         })
         secure.on('error', function (err) {
-          if (self.log.trace()) { self.log.trace({ err: err }, 'error event: %s', new Error().stack) }
+          self.log.trace({ err: err }, 'error event: %s', new Error().stack)
 
           self.emit('error', err)
           sock.destroy()
@@ -866,7 +866,7 @@ Client.prototype.connect = function connect () {
     }
 
     socket.on('data', function onData (data) {
-      if (log.trace) { log.trace('data event: %s', util.inspect(data)) }
+      log.trace('data event: %s', util.inspect(data))
 
       tracker.parser.write(data)
     })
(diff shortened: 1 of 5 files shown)


lib/client/client.js
```diff
@@ -866,7 +866,7 @@ Client.prototype.connect = function connect () {
     }
 
     socket.on('data', function onData (data) {
-      if (log.trace) { log.trace('data event: %s', util.inspect(data)) }
+      log.trace('data event: %s', util.inspect(data))
```
UziTech, line 869: Was the if statement supposed to check for the existence of the function? If the user does not provide a logger does this throw an error?

It looks like some of the if statements check the return value and some check the function (like this one did).

telepsis: Good points. Integration with logging packages isn't terribly strict, though it is functional. The risk is pushed onto the developer that wants to make use of the built-in calls. 

> Was the if statement supposed to check for the existence of the function?

If no log is provided, an `abstract-logger` instance is used to fill out the expected methods. These express nothing and return void, but do prevent errors.

> If the user does not provide a logger (or a logger that does not provide a trace method) does this throw an error?

There is no guarantee or test that a provided logging instance will have the expected methods or that its methods will have the expected signatures/behavior. That does put pressure on developers to use compatible loggers, but that will be the case with or without this PR.

> It looks like some of the if statements check the return value and some check the function (like this one did).

Note that there's an inconsistency in how calls on the log object are being made. In some cases there is no 'if' check at all. 

The single `log.trace` check (vs. `log.trace()`) happens in a type that inherits the same log object used elsewhere, so that appears to be an inconsistency as well.

UziTech: > There is no guarantee or test that a provided logging instance will have the expected methods or that its methods will have the expected signatures/behavior.

On the one hand this could be a problem in theory, on the other hand this hasn't been a problem yet so maybe it isn't in practice.

> Note that there's an inconsistency in how calls on the log object are being made. In some cases there is no 'if' check at all.

Yes making it consistent seems like a good reason for this PR.

UziTech (approved):

jsumners (approved):
LGTM

jsumners:
@UziTech when you get a chance, test out your new publishing privileges with a new patch release (2.2.1).

jsumners:

when opening a new issue.

PR #686, opened 2020-12-02
Fix crash in res.send()

gramakri:
In commit c6fa25985b078ce0ba27da50d1d8676d3e18fe30, "var" usage
was replaced with const. The code previously worked because "var"
has function scoping unlike const which has block scoping.

TypeError: Cannot read property 'attributes' of null

### DIFF
--- a/lib/messages/search_response.js
+++ b/lib/messages/search_response.js
@@ -42,7 +42,7 @@ SearchResponse.prototype.send = function (entry, nofiltering) {
   const self = this
 
   const savedAttrs = {}
-  const save = null
+  let save = null
   if (entry instanceof SearchEntry || entry instanceof SearchReference) {
     if (!entry.messageID) { entry.messageID = this.messageID }
     if (entry.messageID !== this.messageID) { throw new Error('SearchEntry messageID mismatch') }
@@ -66,7 +66,7 @@ SearchResponse.prototype.send = function (entry, nofiltering) {
       }
     })
 
-    const save = entry
+    save = entry
     entry = new SearchEntry({
       objectName: typeof (save.dn) === 'string' ? parseDN(save.dn) : save.dn,
       messageID: self.messageID,


gramakri:
The backtrace we got is
```
2020-12-01T17:39:21.517Z box:ldap TypeError: Cannot read property 'attributes' of null
    at /home/yellowtent/box/node_modules/ldapjs/lib/messages/search_response.js:100:12
    at Array.forEach (<anonymous>)
    at SearchResponse.send (/home/yellowtent/box/node_modules/ldapjs/lib/messages/search_response.js:99:29)
    at /home/yellowtent/box/src/ldap.js:113:17
    at Array.forEach (<anonymous>)
    at finalSend (/home/yellowtent/box/src/ldap.js:112:17)
    at /home/yellowtent/box/src/ldap.js:370:17
    at /home/yellowtent/box/node_modules/async/dist/async.js:473:16
    at replenish (/home/yellowtent/box/node_modules/async/dist/async.js:1006:25)
    at iterateeCallback (/home/yellowtent/box/node_modules/async/dist/async.js:995:17) %s failure to write message %j 172.18.0.7:57498 { messageID: 1075,
  protocolOp: 'LDAPResult',
  status: 0,
  matchedDN: '',
  errorMessage: '',
  referrals: [],
  controls: [] }
```

UziTech:
I wonder if there are more cases of variable shadowing gone wrong. We should add a rule to the linter to prevent this.

jsumners:
This is what I was afraid of in https://github.com/ldapjs/node-ldapjs/pull/674.

jsumners (approved):
I really want to ask for test coverage here to prevent a regression, but I'm unclear how feasible that would be.

UziTech:
I did a search with eslint and only found this shadowing to be a problem. I would prefer to move to eslint and use eslint-config-standard instead of using standard so we can add the no-shadow rule.

During that search I also found that the callback on `getEntryChangeNotificationControl` is never used. I'm not sure if that is a problem.

https://github.com/ldapjs/node-ldapjs/blob/da358a814c6001831807159f825e5ed877dbf899/lib/persistent_search.js#L80-L96

jsumners:
> I would prefer to move to eslint and use eslint-config-standard instead of using standard so we can add the no-shadow rule.

🤔 the point of using `standard` though is that we shouldn't have to keep playing the "update rules" game. We can discuss in a separate issue.



> During that search I also found that the callback on `getEntryChangeNotificationControl` is never used. I'm not sure if that is a problem.

I don't know if it is, but I'd just leave it. One day someone will work up the constitution to clean this whole mess up and these things will get resolved.

jsumners:

when opening a new issue.

PR #688, opened 2020-12-06
use eslint with standard config

UziTech:
This still uses standard for style rules but allows us to add other eslint rules to prevent errors like #686

### DIFF
--- a/lib/client/client.js
+++ b/lib/client/client.js
@@ -697,7 +697,7 @@ Client.prototype.starttls = function starttls (options,
       self._starttls = null
       callback(err)
     })
-    emitter.on('end', function (res) {
+    emitter.on('end', function (_res) {
       const sock = self._socket
       /*
        * Unplumb socket data during SSL negotiation.
@@ -938,7 +938,7 @@ Client.prototype.connect = function connect () {
         f(basicClient, callback)
       },
       inputs: self.listeners('setup')
-    }, function (err, res) {
+    }, function (err, _res) {
       if (err) {
         self.emit('setupError', err)
       }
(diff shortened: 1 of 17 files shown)


lib/client/client.js
```diff
@@ -850,9 +850,9 @@ Client.prototype.connect = function connect () {
   }
 
   // Initialize socket events and LDAP parser.
-  function initSocket (url) {
+  function initSocket (server) {
```
jsumners, line 853: This is my gripe with the "no shadow" rule. This isn't really a shadowing since it is defining a new closure. And in this case, "server" isn't accurate I do not believe. Isn't this parameter literally an instance of `URL`?

jsumners (commented):
In general, this is looking good. I have a question about one of the changes. Also, can we make this no shadow rule ignore `**/*.test.js` files?

lib/client/client.js
UziTech: I called it `server` because that is what the variable is called that is passed in when this function is invoked. This is still shadowing because it is using a variable name from a parent scope.

UziTech:
I turned the no-shadow rule off for tests.

jsumners (approved):
LGTM

UziTech:
I removed the unused variables or set it to allow args and catch blocks to include unused variables if they are prefixed with an underscore.

jsumners:
I see. Hopefully that's all that is needed. This seems to already be getting into the rule maintenance nightmare loop.

UziTech:
That is all I think we need. The eslint rules should be "set and forget", unless we are adding new rules we should never need to change them.

jsumners:

when opening a new issue.

PR #690, opened 2020-12-10
Update guide.md - broken tutorial 

maltewirz:
Dear authors, many thanks for your project. When following your guide, the code crashed when i tried to add a user from the ldif file. I needed the code to change as seen above (the object includes a "attrs" key).

### DIFF
--- a/docs/guide.md
+++ b/docs/guide.md
@@ -403,7 +403,7 @@ the following code in as another handler (you'll need a
       if (!req.dn.rdns[0].attrs.cn)
         return next(new ldap.ConstraintViolationError('cn required'));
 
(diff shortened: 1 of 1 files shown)


maltewirz:
I am currently unable to update this request, but the line "if (!req.dn.rdns[0].attrs.cn || !req.users[req.dn.rdns[0].attrs.cn.value])" also needed to change - the "value" had to be adressed. I will try to update this PR tomorrow.

UziTech (approved):
I think i got all of the `value` places updated. Thanks for this PR :100:

jsumners (approved):
LGTM

Issue #679, opened 2020-11-13
change ldapjs.org dns record


UziTech:
The `ldapjs.org` dns needs to be updated. I updated the gh-pages and got this error by email:

```
The custom domain for your GitHub Pages site is pointed at an outdated IP address.
You must update your site's DNS records if you'd like it to be available via your custom domain.
```

It looks like the dns should point to

```
185.199.108.153
185.199.109.153
185.199.110.153
185.199.111.153
```

@jsumners do you know who controls the domain name?

jsumners:
Ugh. That's a thing we never got around to dealing with. Technically Joyent still owns all of that. I'll reach out to people and see if we can get it moved somewhere else. But it may be that we just need to kill of the domain.

UziTech:
We could just use `ldapjs.github.io` by moving the gh-pages branch to the repo `ldapjs/ldapjs.github.io`

jsumners:
Yeah, that sounds like a good solution.

UziTech:
I created the repo and set up https://ldapjs.github.io/ I think it would be better if we could work something out with Joyent to keep the domain, but if not we will just need to point everything to the new domain.

jsumners:
@melloc do you have any opinions about managing DNS for `ldapjs.org`?

Personally, I'm of the opinion that we should just let the domain die. Maybe add a year to it and set it up as a 301 to `ldapjs.github.io`. It'll be far easier to maintain as an organization going forward. If we were to try to keep `ldapjs.org` then someone would have to keep it paid for, and the org will always be wondering if that person will keep up with it or just disappear and be stuck without the domain.

jsumners:
👋

On February 22, 2023, we released [version 3](https://github.com/ldapjs/node-ldapjs/releases/tag/v3.0.0) of this library. As a result, we are closing this issue/pull request.

Please see [issue #839](https://github.com/ldapjs/node-ldapjs/issues/839) for more information, including how to proceed if you feel this closure is in error.

PR #693, opened 2020-12-13
deploy docs to gh-pages

UziTech:
- Add `npm run docs` which build the docs into `./public`
- Add GitHub Action to deploy docs on push to master


.github/workflows/docs.yml
```diff
@@ -0,0 +1,22 @@
...
+    runs-on: ubuntu-latest
+    steps:
+      - uses: actions/checkout@v2
+      - uses: actions/setup-node@v1
+      - name: Install Packages
+        run: npm install
+      - name: Build Docs
+        run: npm run docs
+      - name: Deploy 🚢
+        uses: peaceiris/actions-gh-pages@v3
```
jsumners, line 20: I'd really rather publish to a `ldapjs.github.io` repository instead of a `gh-pages` branch.

I have this action that does this for Hugo based projects -- https://github.com/jsumners/gh-action-hugo-deploy

It'd be easy to rewrite for this setup since we probably wouldn't even need Docker.

build-docs.js
```diff
@@ -0,0 +1,101 @@
+const fs = require('fs/promises')
```
jsumners, line 1: Would you be opposed to moving this to a root level `scripts` directory?

build-docs.js
```diff
@@ -0,0 +1,101 @@
...
+  const dest = path.join(dist, 'media')
+  const src = path.join(docs, 'branding', 'media')
+  await fs.mkdir(dest)
+  await fs.mkdir(path.join(dest, 'css'))
+  await fs.mkdir(path.join(dest, 'img'))
+  await fs.copyFile(path.join(src, 'css', 'style.css'), path.join(dest, 'css', 'style.css'))
+  await fs.copyFile(path.join(src, 'img', 'logo.svg'), path.join(dest, 'img', 'logo.svg'))
+}
+
+createDocs()
```
jsumners, line 101:
```suggestion
createDocs().then(() => {}).catch(console.error)
```

Could be better improved to return an exit code if this is worked directly into a GH Action via https://github.com/actions/github-script

docs/client.md
```diff
@@ -1,13 +1,16 @@
 ---
 title: Client API | ldapjs
-markdown2extras: tables
 ---
 
 # ldapjs Client API
 
+<div class="intro">
```
jsumners, line 7: Do we really need to introduce structural HTML into the Markdown?

UziTech: I was just trying to keep the style the same as it is now. We can change the style however we want. We should probably also remove the Joyent logo at some point. But this PR was mostly to get the website up to date with the docs changes. We can change the style in another PR.

.github/workflows/docs.yml
UziTech: Does that action have to use hugo? We could change it to that in another PR. We should move the CNAME to that repo then since most links still go to `ldapjs.org` instead of `ldapjs.github.io`.

We still need to figure out #679.

jsumners: I'm suggesting we create a new action in the same vein. That one is very much specific to Hugo.

For now, I've gone ahead and removed the custom domain from this repository.

I know. I just updated #679 with at least a little bit of movement.

docs/client.md
jsumners: I'm not completely opposed. Just wondering if we think it is worth the noise in the docs.

.github/workflows/docs.yml
UziTech: Looks like there are quite few actions that do that. The best one I am finding is [s0/git-publish-subdir-action](https://github.com/marketplace/actions/push-git-subdirectory-as-branch)

docs/client.md
UziTech: I think it is worth it for now but I'm ok with changing it in another PR.

.github/workflows/docs.yml
jsumners: Looks good to me. Seems it can be run totally within the action so we do not need to commit build artifacts to this repository.

jsumners (approved):
image

.github/workflows/docs.yml
UziTech: I used cpina/github-action-push-to-another-repository instead but we will need to add a secret with repo persmission to `API_TOKEN_GITHUB`

jsumners: Done.

PR #704, opened 2021-02-19
Emit error event if other error events are not listened to

UziTech:
Emit error event if other error events are not listened to.

Currently if someone wants to catch client errors they have to subscribe to multiple error events

```js
client.on("error", (err) => {
  assert.ifError(err);
});
client.on("setupError", (err) => {
  assert.ifError(err);
});
client.on("connectError", (err) => {
  assert.ifError(err);
});
client.on("connectTimeout", (err) => {
  assert.ifError(err);
});
client.on("connectRefused", (err) => {
  assert.ifError(err);
});
```

This PR will allow them to just subscribe to the `error` event to catch all of them but still allow them to subscribe to more if they want to.

### TODO:

- [ ] Add Tests

### DIFF
--- a/docs/client.md
+++ b/docs/client.md
@@ -20,6 +20,10 @@ The code to create a new client looks like:
       url: ['ldap://127.0.0.1:1389', 'ldap://127.0.0.2:1389']
     });
 
+    client.on('error', (err) => {
+      // handle connection error
+    })
+
 You can use `ldap://` or `ldaps://`; the latter would connect over SSL (note
 that this will not use the LDAP TLS extended operation, but literally an SSL
 connection to port 636, as in LDAP v2). The full set of options to create a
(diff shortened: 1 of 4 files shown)


UziTech:
@jsumners I wanted to get your input on this before I go further. I wasn't sure if this would be too much of a breaking change. If they weren't listening to any errors before a `connectRefused`, for example, would just fail silently. With this PR if there are no listeners for the `error` or `connectRefused` event than the node process will crash because of the error (Which I think is the right thing to do anyway).

jsumners:
Seems fair to me. I could argue that this "fixes" a hole in the error handling and therefore isn't a major change.

jsumners (approved):
LGTM

Should we try to squeeze some docs in somewhere?

UziTech:
It looks like only `connectTimeout` and `connectRefused` don't emit an error already if the aren't listened to. The other errors (`setupError` and `connectError`) also emit an `error` as well as the specific event.

jsumners:
Ship it.
giphy

UziTech:
Ya, I will add docs to this before merging.

UziTech:
@jsumners what do you think about the added [docs](https://github.com/ldapjs/node-ldapjs/pull/704/files#diff-5323dcc2b5e999ce4ad5b916d9f63514e20a8d40d86ff16d89a1c21aad7179f0)?

jsumners:
LGTM

jsumners:

when opening a new issue.

PR #708, opened 2021-02-25
docs: simplify docs

UziTech:
- remove unused js and css
- update repo information
- simplify build-docs script


.gitignore
```diff
@@ -77,4 +77,4 @@ xcuserdata
 *.env
 
 # built docs
-public
+/public
```
jsumners, line 80: Should it be? 
```suggestion
public/
```

`/public` makes me think it is going to try to ignore a "public" directory at the root of the file system, not the project.

UziTech: No `public/` will make it ignore any folder named `public` in the project. `/public` is the only way to make it only ignore a folder named `public` in the root of the project.

jsumners: 👍

jsumners:

when opening a new issue.

PR #711, opened 2021-03-15
object to Attribute conversion on client should not call .toString()

joaopbnogueira:
**Bug:**
- When trying to add a binary element, such as `jpegPhoto` of type `Buffer`, the client's add method will serialize it when converting a user object to an array of attributes, which works fine for most cases, but not for binary data on Buffers.

The following pseudo-code results in an invalid image on an LDAP server
```
client.add(dn, {jpegPhoto: (buffer)});
```

Adding as an attribute works as expected:
```
const attr = new Attribute({type: 'jpegPhoto'});
attr.addValue(buffer);

client.add(dn, [attr]);
```

### DIFF
--- a/lib/client/client.js
+++ b/lib/client/client.js
@@ -236,10 +236,10 @@ Client.prototype.add = function add (name, entry, controls, callback) {
       const attr = new Attribute({ type: k })
       if (Array.isArray(save[k])) {
         save[k].forEach(function (v) {
-          attr.addValue(v.toString())
+          attr.addValue(v)
         })
       } else {
-        attr.addValue(save[k].toString())
+        attr.addValue(save[k])
       }
       entry.push(attr)
     })


jsumners (changes requested):
Please see lines 207 and 208. The documentation for this method is rather clear.

I'm sure there is room for improvement here, but an accompanying unit test would be required.

joaopbnogueira:
@jsumners  thanks for the review. 

I did not notice the method documentation, but then I guess there's the question: is the documentation the intended behavior, in which case no change should be made to the code, or is the documentation merely clarifying how it works?

UziTech:
This PR passes all of the current tests so I would say that is more of a clarification than a requirement.

UziTech:
I'm not sure how a server would handle something like a boolean or number. Perhaps we should check if it is a `Buffer` and only cast to string if it is not.

jsumners:
> This PR passes all of the current tests so I would say that is more of a clarification than a requirement.

I would argue that our tests do not currently cover the situation. A failing test should be added prior to any change to the code to illustrate the change.

joaopbnogueira:
@Y

> I'm not sure how a server would handle something like a boolean or number. Perhaps we should check if it is a `Buffer` and only cast to string if it is not.

This already happens, but on the Attribute class itself:

```
Attribute.prototype.addValue = function addValue (val) {
  if (Buffer.isBuffer(val)) {
    this._vals.push(val)
  } else {
    this._vals.push(Buffer.from(val + '', _bufferEncoding(this.type)))
  }
}
```

joaopbnogueira:
> > This PR passes all of the current tests so I would say that is more of a clarification than a requirement.
> 
> I would argue that our tests do not currently cover the situation. A failing test should be added prior to any change to the code to illustrate the change.

I'll have a look into adding a test to reproduce the bug.

UziTech:
> This already happens, but on the Attribute class itself

In that case I would say it is a bug that `client.add` casts to string.

UziTech:
I think this is fixed by #754

jsumners:
Given the lack of movement, and the potential fix going out soon, I'm closing this. If this is not resolved by the other pull request, feel free to reopen it when you have time to finish it.

jsumners:

when opening a new issue.

PR #721, opened 2021-05-05
Add missing error emit.

seewer:
This is a small one. The exception is never emitted as an error, so is only seen when traceing. Stumbled over it while debugging something else. 

Signed-off-by: Philippe Seewer <philippe.seewer@bfh.ch>

### DIFF
--- a/lib/client/client.js
+++ b/lib/client/client.js
@@ -1274,6 +1274,7 @@ Client.prototype._sendSocket = function _sendSocket (message,
     if (timer) { clearTimeout(timer) }
 
     log.trace({ err: e }, 'Error writing message to socket')
+    self.emit('error', e)
     return callback(e)
   }
 }


jsumners (commented):
Thank you for the PR. Are you able to cover this with a unit test?

jsumners:
It's not clear to me this should even be emitted. The error is already supplied to the callback. What situation causes this to arise?

seewer:
It was more of a consistency issue. Every other traced error is emitted, just not this one. So my pattern brain automatically kicked in. And it helped while trying to figure out our concurrency issues.

Running the same tests now with the focus on paged searches actually causes a double emit, which is clearly wrong. I'm closing this PR as I can't construct a use case in a production environment where this might be necessary.

jsumners:
Thank you for sticking with it. We know the combination of emitter and callbacks is a big problem. We hope to one day get the time and fortitude to clean it all up.

jsumners:

when opening a new issue.

PR #723, opened 2021-05-07
SearchPager: Implement queueing events until a listener appears

seewer:
This fixes the same problems for paged searches as does
https://github.com/ldapjs/node-ldapjs/pull/610 for unpaged searches.

By passing an EventEmitter via callback there exist cases when events
are emitted before listeners are registered, resulting in missed
events.

The Change turns SearchPager into a CorkedEmitter which is already
used as a solution for non paged searches. Doing so requires the
internal 'search' event to be dropped.

This change adapts a test case originally by László Szűcs (@ifroz).

Signed-off-by: Philippe Seewer <philippe.seewer@bfh.ch>

### DIFF
--- a/lib/client/search_pager.js
+++ b/lib/client/search_pager.js
@@ -10,6 +10,8 @@ const assert = require('assert-plus')
 // var Protocol = require('../protocol')
 const PagedControl = require('../controls/paged_results_control.js')
 
+const CorkedEmitter = require('../corked_emitter.js')
+
 /// --- API
 
 /**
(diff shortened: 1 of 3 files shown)


lib/client/search_pager.js
```diff
@@ -29,19 +31,20 @@ const PagedControl = require('../controls/paged_results_control.js')
  *                will be emitted (and 'end' will not be).  By listening to
  *                'pageError', a successful search that lacks paging will be
  *                able to emit 'end'.
- * 3. search    - Emitted as an internal event to trigger another client search.
  */
 function SearchPager (opts) {
   assert.object(opts)
   assert.func(opts.callback)
   assert.number(opts.pageSize)
+  assert.func(opts.sendRequest)
```
jsumners, line 39: Does this alter the public API and introduce a breaking change?

seewer: The search event was clearly marked as internal and only used in one place.

jsumners:
Looks like this PR branch is not up-to-date with the target `master` branch. Please rebase.

jsumners (approved):

jsumners (approved):
Looks good to me.

PR #727, opened 2021-05-19
connection router to accomodate multithreading

CoryGH:
connection router hook added to accommodate multithreaded servers

e.g. passing `options.connectionRouter(connection)` will allow for:

    function connectionRouter(connection) {
        sendToThread(findRoute(connection), connection);
    }
    
    function sendToThread(threadId, connection) {
        ...wrap in packet for IPC and send to thread...
    }
    
    function threadMessage(message) {
        notListeningServerOnChildThread.newConnection(message.connection);
    }

in order to preserve extant functionality while adding multi-threading support.

### DIFF
--- a/docs/examples.md
+++ b/docs/examples.md
@@ -550,3 +550,76 @@ To test out this example, try:
 $ ldapsearch -H ldap://localhost:389 -x -D cn=demo,dc=example,dc=com \
   -w demo -b "dc=example,dc=com" objectclass=*
 ```
+
+# Multi-threaded Server
+
+This example demonstrates multi-threading via the `cluster` module utilizing a `net` server for initial socket receipt.  An alternate example demonstrating use of the `connectionRouter` `serverOptions` hook is available in the `examples` directory.
+
+```
+const cluster = require('cluster');
+const ldap = require('ldapjs');
+const net = require('net');
+const os = require('os');
+
+const threads = [];
+threads.getNext = function () {
+    return (Math.floor(Math.random() * this.length));
+};
+
+const serverOptions = {
+    port: 1389
+};
+
+if (cluster.isMaster) {
+    const server = net.createServer(serverOptions, (socket) => {
+        socket.pause();
+        console.log('ldapjs client requesting connection');
+        let routeTo = threads.getNext();
+        threads[routeTo].send({ type: 'connection' }, socket);
+    });
+
+    for (let i = 0; i < os.cpus().length; i++) {
+        let thread = cluster.fork({
+            'id': i
+        });
+        thread.id = i;
+        thread.on('message', function (msg) {
+
+        });
+        threads.push(thread);
+    }
+
+    server.listen(serverOptions.port, function () {
+        console.log('ldapjs listening at ldap://127.0.0.1:' + serverOptions.port);
+    });
+} else {
+    const server = ldap.createServer(serverOptions);
+
+    let threadId = process.env.id;
+
+    process.on('message', (msg, socket) => {
+        switch (msg.type) {
+            case 'connection':
+                server.newConnection(socket);
+                socket.resume();
+                console.log('ldapjs client connection accepted on ' + threadId.toString());
+        }
+    });
+
+    server.search('dc=example', function (req, res, next) {
+        console.log('ldapjs search initiated on ' + threadId.toString());
+        var obj = {
+            dn: req.dn.toString(),
+            attributes: {
+                objectclass: ['organization', 'top'],
+                o: 'example'
+            }
+        };
+
+        if (req.filter.matches(obj.attributes))
+            res.send(obj);
+
+        res.end();
+    });
+}
+```
--- a/docs/examples.md
+++ b/docs/examples.md
@@ -555,7 +555,7 @@ $ ldapsearch -H ldap://localhost:389 -x -D cn=demo,dc=example,dc=com \
 
 This example demonstrates multi-threading via the `cluster` module utilizing a `net` server for initial socket receipt.  An alternate example demonstrating use of the `connectionRouter` `serverOptions` hook is available in the `examples` directory.
 
-```
+```js
 const cluster = require('cluster');
 const ldap = require('ldapjs');
 const net = require('net');
--- a/lib/server.js
+++ b/lib/server.js
@@ -314,6 +314,8 @@ function Server (options) {
   }
 
   self.newConnection = function (conn) {
+    // TODO: make `newConnection` available on the `Server` prototype
+    // https://github.com/ldapjs/node-ldapjs/pull/727/files#r636572294
     setupConnection(conn)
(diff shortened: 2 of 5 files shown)


UziTech (changes requested):
Can you add a test and update docs?

CoryGH:
> 
> 
> Can you add a test and update docs?

Sure, I think I have a handle on your test system.

UziTech:
it seems like multi threading with the `connectionRouter` option would only make use of ldapjs to start the server.

It seems like it might be easier to start the server and send connection to worker threads without ldapjs.

```js
const net = require("net");
net.createServer(connectionRouter)
```

What are the benefits of using ldapjs to start the server when you are using `notListeningServerOnChildThread` to do all of the actual work?

CoryGH:
> What are the benefits of using ldapjs to start the server when you are using `notListeningServerOnChildThread` to do all of the actual ldap work?

My aim was to make things minimally-invasive from an API standpoint.  The `connectionRouter` function is called at the master thread level to tell the thread-manager (whichever it may be) to forward to a child thread which must then call `newConnection(connection)` in order to initialize the ldapjs side for the connection.  It's designed around a forking design pattern but is extensible to cross-machine threading engines as well.

The `server.listen` function looks lightweight enough to do what you're suggesting (if what you mean is to create a net/tls server outside of ldapjs then use `server.newConnection(connection)` to leverage the ldapjs functionality with the socket (the change to make `server.newConnection(connection)` publicly-visible would still be needed, the `connectionRouter` piece wouldn't be needed at that point.  Generally I shy away from moving entire members outside of libraries wholesale like that (in this case reinforced when I was reviewing `server.newConnection(connection)` because there's a lot of library-specific code in there leveraging the this/self reference to set the appropriate state and there's no real guarantee some of that logic or things similar to it from the patterns in the code won't get moved to the `listen` function in the future (that seems like a bigger commitment to me on the part of library developers/maintainers than adding a hook, though I could be wrong.)

CoryGH:
This is a mess in trying to get around the tap test framework's thread handling, here's what I have thus far though - might need to take a break from it today but the code paths are being hit (failing test with "test unfinished" and I haven't pinned down why as of yet (I placed this in `server.test.js`, had to wrap the other tests in `if (cluster.isMaster) { ... }` as well to avoid them running on child threads.

    tap.test('multithreading', function (t) {
      let threads = []
      const server = ldap.createServer({
        connectionRouter: (connection) => {
          if (cluster.isMaster) {
            threads[0].send({
              type: 'connection'
            }, connection)
          }
        }
      })
      if (cluster.isMaster) {
        let clientHolder = { client: null }
        for (let i = 0; i < 1; i++) {
          let thread = cluster.fork({
            id: i
          })
          thread.on('message', function (msg, connection) {
            switch (msg.type) {
              case 'connected':
                clientHolder.client.unbind()
                threads[0].disconnect()
                t.ok(true, 'client connected')
                t.end()
                server.close(() => t.end())
                break
            }
          })
          threads.push(thread)
        }
        t.ok(server)
        server.listen(5555, 'localhost', function () {
          t.ok(true, 'server listening on ' + server.url)
    
          server.getConnections(function (err, count) {
            t.error(err)
            t.equal(count, 0)
            setTimeout(() => {
              clientHolder.client = ldap.createClient({ port: 5555, url: server.url })
              clientHolder.client.on('connect', function () { })
            }, 1000)
          })
        })
      } else {
        process.on('message', (msg, connection) => {
          switch (msg.type) {
            case 'connection':
              server.newConnection(connection)
              process.send({
                type: 'connected'
              })
              break
          }
        })
      }
    })

CoryGH:
I would agree that if the `server.listen()` call can be guaranteed not to include state-specific bindings as the library evolves it would be better to just make `server.newConnection(connection)` publicly-visible without the `connectionRouter(connection)` hook.

UziTech:
I think it would be easiest to mock the threads instead of trying to run the tests in actual threads. Just testing that `connectionRouter` option does what it is supposed to and that `newConnection` is public should work.

CoryGH:
> 
> 
> I think it would be easiest to mock the threads instead of trying to run the tests in actual threads. Just testing that `connectionRouter` option is called when it is supposed to be and that `newConnection` is public should work.

Would you like it with or without `connectionRouter` given the ability to bypass the listen function?

UziTech:
> Would you like it with or without `connectionRouter` given the ability to bypass the listen function?

Maybe two tests, on each way?

CoryGH:
> 
> 
> > Would you like it with or without `connectionRouter` given the ability to bypass the listen function?
> 
> Maybe two tests, one each way?

Tests added and connectionRouter hook added back in.

CoryGH:
@UziTech updated to fix linting issue

lib/server.js
```diff
@@ -313,7 +313,7 @@ function Server (options) {
     return c
   }
 
-  function newConnection (conn) {
+  self.newConnection = function (conn) {
```
jsumners, line 316: Why does this need to be `self.newConnection`?

jsumners (commented):
Are you able to add some documentation. An example in the docs would be quite helpful.

lib/server.js
UziTech: It just needs to be a public function. We should probably move it to the prototype.

CoryGH: I can throw together a simple example for the documentation.

CoryGH:
Added examples in the examples directory for both multithreading via the `connectionRouter` hook in conjunction with `newConnection(connection)` and another by replacing the whole listen server to use `newConnection(connection)` for the node-ldapjs bindings.

docs/examples.md
```diff
@@ -550,3 +550,76 @@ To test out this example, try:
 $ ldapsearch -H ldap://localhost:389 -x -D cn=demo,dc=example,dc=com \
   -w demo -b "dc=example,dc=com" objectclass=*
 ```
+
+# Multi-threaded Server
+
+This example demonstrates multi-threading via the `cluster` module utilizing a `net` server for initial socket receipt.  An alternate example demonstrating use of the `connectionRouter` `serverOptions` hook is available in the `examples` directory.
+
+```
```
jsumners, line 558:
```suggestion
```js
```

lib/server.js
```diff
@@ -313,7 +313,7 @@ function Server (options) {
...
     setupConnection(conn)
```
jsumners, line 317:
```suggestion
    // TODO: make `newConnection` available on the `Server` prototype
    // https://github.com/ldapjs/node-ldapjs/pull/727/files#r636572294
    setupConnection(conn)
```

We'll solve this at a later date.

jsumners (approved):
Looks good to me once CI is green.

UziTech:
@CoryGH you can run `npm run lint` to fix the linting issues.

jsumners:

when opening a new issue.

PR #743, opened 2021-07-08
Update url.js

MSZ006ZetaCY:
to fix "url.URL is not a constructor" in new nodejs api

### DIFF
--- a/lib/url.js
+++ b/lib/url.js
@@ -10,7 +10,11 @@ module.exports = {
   parse: function (urlStr, parseDN) {
     let parsedURL
     try {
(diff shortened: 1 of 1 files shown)


lib/url.js
```diff
@@ -10,7 +10,11 @@ module.exports = {
   parse: function (urlStr, parseDN) {
     let parsedURL
     try {
-      parsedURL = new url.URL(urlStr)
+      if(url.URL){
```
MSZ006ZetaCY, line 13: if (url.URL) {

jsumners:

when opening a new issue.

PR #744, opened 2021-07-08
Update url.js

MSZ006ZetaCY:
fix "url.URL is not a constructor" problem


jsumners (changes requested):
Under what condition is this necessary? Where is the unit test covering this change?

MSZ006ZetaCY:
I‘m so sorry.
Now I known this isn't run in browser.

jsumners:

when opening a new issue.

Issue #748, opened 2021-07-10
Anonymous Bind DN Inconsistencies


tpretz:
In server.js as part of connection setup the bindDN is defaulted to cn=anonymous.

However, once an explicit anonymous bind has executed, this is replaced with a DN object with no internal RDNs (so, just "")

It would be nice if a consistent value could be used to represent an anonymous user.

In my case i just did the following to override the name attribute in the handling of anon binds, but this could be adjusted in any number of places, or equally the "" value used in both cases.

```
// check anonymous bind
  if (req.protocolOp === Protocol.LDAP_REQ_BIND &&
      req.dn.toString() === '' &&
      req.credentials === '') {

      req.name = new DN([new dn.RDN({cn: 'anonymous'})]);
    return {
      backend: self,
      handlers: [defaultNoOpHandler]
    }
  }
```

jsumners:
What is this issue in reference to?

tpretz:
It is in reference to lib/server.js and how it handles bind and unbind actions.

A fresh connection begins with a bindDN of cn=anonymous 
https://github.com/ldapjs/node-ldapjs/blob/3430f05d91bc152b18febe1f04665c8aea0faf11/lib/server.js#L285

When a successful bind is performed, bindDN is replaced with the contents of that bind request.
https://github.com/ldapjs/node-ldapjs/blob/3430f05d91bc152b18febe1f04665c8aea0faf11/lib/server.js#L409

For an anonymous bind, the result of that is a 0 length string, which is inconsitent with the implementation default of cn=anonymous (they should be aligned, one way or the other)

It is understood that server.js is maintaining the variable connection.ldap.bindDN (based on the success of bind callbacks). 
I did also notice no similar handing for unbind actions, which should set bindDN back to the anonymous value

jsumners:
Would you like to send a Pull Request to address this issue? Remember to add unit tests.

tpretz:
sure, ill put something together, unless you prefer an emtpy rdn representing an anon bind ill stick with cn=anonymous. 
As i assume that was the original intended value.

jsumners:
⚠️ This issue has been locked due to age. If you have encountered a recent
problem that seems to be covered by this issue, please open a new issue.

Please include a [minimal reproducible example](https://en.wikipedia.org/wiki/Minimal_reproducible_example)
when opening a new issue.

PR #750, opened 2021-07-12
Server - Consistent BindDN

tpretz:
Ensuring a connection in an anonymous state has a consistent bindDN variable, regardless of being implicit, or explicitly bound.

Added hook to switch to anonymous on unbind too.
Server should close such connections but added as a safeguard to ensure permissions are dropped regardless of connection termination state.

https://github.com/ldapjs/node-ldapjs/issues/748


UziTech (approved):
LGTM

jsumners (approved):
Looks good to me.

jsumners:

when opening a new issue.

PR #751, opened 2021-07-14
OpenLDAP Syncrepl support

tpretz:
It can be useful to be able to interact with the OpenLDAP replication standard, both as a client and as a server.

This PR will add support for the various controls, and messages that are required


lib/controls/sync_done_control.js
```diff
@@ -0,0 +1,88 @@
+// Copyright 2011 Mark Cavage, Inc.  All rights reserved.
+
```
jsumners, line 2: These new files should not start with this outdated copyright notice. They should also start with `'use strict'`.
```suggestion
'use strict'

```

UziTech:
Looks like this just needs some tests.

clayrisser:
Also, how would I use this? Is there a basic example?

UziTech:
This just needs some tests in order to be merged in. @tpretz are you still working on this?

clayrisser:
@tpretz  do you have an example on how to use this?

jsumners:
@clayrisser unfortunately, I think this PR is abandoned. You can take it up by following https://gist.github.com/jsumners/461ef7a64545108635cc437fde112721

tpretz:
unfortunately i have not had time recently to work on the project that was pushing these additions

When i do get back to it i'll raise again, or if anyone else wants to look at / enhance my source branch will remain.

jsumners:

when opening a new issue.

PR #753, opened 2021-07-29
fix: modifyDn length fix

### LINKED ISSUE #740: client.modifyDN fails
client.modifyDN fails without any error message when dn is more than 112 characters

JohnTJohnston:
This should resolve #740.

### DIFF
--- a/lib/messages/moddn_request.js
+++ b/lib/messages/moddn_request.js
@@ -60,7 +60,7 @@ ModifyDNRequest.prototype._toBer = function (ber) {
     const len = Buffer.byteLength(s)
 
     ber.writeByte(0x80) // MODIFY_DN_REQUEST_NEW_SUPERIOR_TAG
-    ber.writeByte(len)
+    ber.writeLength(len)
     ber._ensure(len)
     ber._buf.write(s, ber._offset)
     ber._offset += len
--- a/test/client.test.js
+++ b/test/client.test.js
@@ -694,6 +694,34 @@ tap.test('modify DN excessive length (GH-480)', function (t) {
   })
 })
 
+tap.test('modify DN excessive superior length', function (t) {
+  const { BerReader, BerWriter } = require('asn1')
+  const ModifyDNRequest = require('../lib/messages/moddn_request')
+  const ber = new BerWriter()
+  const entry = 'cn=Test     User,ou=A Long OU                  ,ou=Another Long OU                ,ou=Another Long OU              ,dc=acompany,DC=io'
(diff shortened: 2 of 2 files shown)


UziTech:
Can you write a test for this?

JohnTJohnston:
This should repro testing against AD. In my case, the new superior was 0x86 bytes long, and wireshark ldap disector flagged the length as illegal.  Refer to "BER Lengths" in [https://ldap.com/ldapv3-wire-protocol-reference-asn1-ber/](url)

```
const cn = "cn=existing user";
const oldDn = cn + ",ou=users,dc=aCompany,dc=com";
const newDn = cn + ",someLongNewSuperiorGreaterThan0x80Lengh";

client.modifyDN(oldDn, newDn, (err) => {
    if (err) console.log(err);
    else console.log("ok");
});
```

Running above will log "ok", but the object is not moved.

jsumners (changes requested):
We need at least an integration test in the same vein as https://github.com/ldapjs/node-ldapjs/blob/91de94d1a625b66b753091697632cbc6f433913d/test-integration/client/issues.test.js#L12-L41

jsumners (changes requested):
We need at least an integration test in the same vein as https://github.com/ldapjs/node-ldapjs/blob/91de94d1a625b66b753091697632cbc6f433913d/test-integration/client/issues.test.js#L12-L41

JohnTJohnston:
That won't work.  The new superior must be long.  A long RDN will not repro the problem.  So the test can only be written if your test directory has an existing long superior, which in my case was an OU.

JohnTJohnston:
Perhaps I'm missing the point of your unit/integration tests.  Is it essential to reproduce a failure prior to the fix?  Is it possible for unit tests to get access to the output BER stream to confirm that the length was inserted  correctly?

jsumners:
> That won't work. The new superior must be long. A long RDN will not repro the problem. So the test can only be written if your test directory has an existing long superior, which in my case was an OU.

I am showing an example of a test for a _similar_ problem. One would need to be written to cover this problem.

> Perhaps I'm missing the point of your unit/integration tests. Is it essential to reproduce a failure prior to the fix? Is it possible for unit tests to get access to the output BER stream to confirm that the length was inserted correctly?

Yes. It is essential. Otherwise we have no proof that the code does what it is purported to do. It also helps prevent changes in the future that would re-introduce the problem.

UziTech (approved):
LGTM

jsumners (approved):
Fantastic! Everything checks out locally. Thank you for providing the test.

Looks good to me.

jsumners:

when opening a new issue.

PR #754, opened 2021-07-29
fix: client.add not adding buffers properly

JohnTJohnston:
Fixed while getting AD thumbnailPhoto to work.

To repro, add a thumbnailPhoto to AD.

Note: I don't believe I ever reported this problem, but I've been using the fix for a long time.

### DIFF
--- a/lib/client/client.js
+++ b/lib/client/client.js
@@ -238,6 +238,8 @@ Client.prototype.add = function add (name, entry, controls, callback) {
         save[k].forEach(function (v) {
           attr.addValue(v.toString())
         })
+      } else if (Buffer.isBuffer(save[k])) {
+        attr.addValue(save[k]);
       } else {
         attr.addValue(save[k].toString())
       }
--- a/lib/client/client.js
+++ b/lib/client/client.js
@@ -239,7 +239,7 @@ Client.prototype.add = function add (name, entry, controls, callback) {
           attr.addValue(v.toString())
         })
       } else if (Buffer.isBuffer(save[k])) {
-        attr.addValue(save[k]);
+        attr.addValue(save[k])
       } else {
         attr.addValue(save[k].toString())
       }
--- a/test/client.test.js
+++ b/test/client.test.js
@@ -508,6 +508,40 @@ tap.test('add success with object', function (t) {
   })
 })
 
+tap.test('add buffer', function (t) {
+  const { BerReader } = require('asn1')
+  const dn = `cn=add, ${SUFFIX}`
+  const attribute = 'thumbnailPhoto'
+  const binary = 0xa5
+  const entry = {
+    [attribute]: Buffer.from([binary])
+  }
+  const write = t.context.client._socket.write
+  t.context.client._socket.write = (data, encoding, cb) => {
+    const reader = new BerReader(data)
+    t.equal(data.byteLength, 49)
+    t.ok(reader.readSequence())
+    t.equal(reader.readInt(), 0x1)
+    t.equal(reader.readSequence(), 0x68)
+    t.equal(reader.readString(), dn)
+    t.ok(reader.readSequence())
+    t.ok(reader.readSequence())
+    t.equal(reader.readString(), attribute)
+    t.equal(reader.readSequence(), 0x31)
+    t.equal(reader.readByte(), 0x4)
+    t.equal(reader.readByte(), 1)
+    t.equal(reader.readByte(), binary)
+    t.context.client._socket.write = write
+    t.context.client._socket.write(data, encoding, cb)
+  }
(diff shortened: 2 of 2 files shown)


UziTech (changes requested):
Can you add a test?

JohnTJohnston:
Sorry, I must have missed something in your documentation.  I don't know what is assumed by your test runs.  I need to be able to add a thumbnail photo to AD.  Do you run your tests against AD?  Can I count on the existence of a particular user?

lib/client/client.js
```diff
@@ -238,6 +238,8 @@ Client.prototype.add = function add (name, entry, controls, callback) {
         save[k].forEach(function (v) {
           attr.addValue(v.toString())
         })
+      } else if (Buffer.isBuffer(save[k])) {
+        attr.addValue(save[k]);
```
jsumners, line 242: Please follow the linting rules. Use `npm run lint`.

jsumners (changes requested):
> To repro, add a thumbnailPhoto to AD.

Sorry. This is not an effective means of proving the change. First, not everyone has access to an Active Directory instance. Second, it does not prevent regressions in future releases. A unit test is required to satisfy these requirements.

JohnTJohnston:
Sorry, I'm not clear what is expected of your unit tests.  How are they supposed to work without a back end directory to verify functionality?  Not sure what other attributes require passing a buffer into add, so how to test?

Seems like I may have to live with my private patches...

jsumners:
> Sorry, I'm not clear what is expected of your unit tests. How are they supposed to work without a back end directory to verify functionality? Not sure what other attributes require passing a buffer into add, so how to test?
> 
> Seems like I may have to live with my private patches...

The client unit tests are in https://github.com/ldapjs/node-ldapjs/blob/91de94d1a625b66b753091697632cbc6f433913d/test/client.test.js

We further have "integration" tests at https://github.com/ldapjs/node-ldapjs/tree/91de94d1a625b66b753091697632cbc6f433913d/test-integration/client

The integration tests use an standard OpenLDAP instance as built by https://github.com/ldapjs/docker-test-openldap and provided via https://github.com/ldapjs/docker-test-openldap/pkgs/container/docker-test-openldap/openldap

The integration tests are useful in cases where writing a unit test is either too difficult or not sufficient.

Note: I need to do some work to update the Docker repository to GitHub's latest, truly anonymous, access mechanisms. But it is attainable currently through authenticated means.

lib/client/client.js
JohnTJohnston: npm run lint completed

note: I have eslint installed globally, but it was not working.  So I installed it as a local dev dependency along with required peers pulled in by extends standard: eslint-plugin-import eslint-plugin-node, and eslint-plugin-promise.

Please advise as to best practices here.  Seems like these should be included as dev dependencies in the repo.

JohnTJohnston:
I am not familiar with openldap attributes.  this bug fix is specfic to attributes that require a javascript buffer to be passed in, and for all I know, openldap might have been smart enough not to have such attributes.

lib/client/client.js
jsumners: `npm install` should install all required dev dependencies. I know we were using vanilla `standard` but that switched to support some configuration extension. Maybe we missed adding some dependencies @UziTech ?

jsumners:
> I am not familiar with openldap attributes. this bug fix is specfic to attributes that require a javascript buffer to be passed in, and for all I know, openldap might have been smart enough not to have such attributes.

We really can't merge patches without a test to cover their functionality. This PR addresses reading of binary attributes, yes? My initial theory would be that updating the server embedded in the unit tests to return an entry that has a `Buffer` valued attribute would allow for a unit test to be written to cover this change.

lib/client/client.js
JohnTJohnston: No, it's there.  Not sure what my problem was.  Sorry for this noise.

JohnTJohnston:
This was necessary to modify a binary attribute.  This following test passes, but it passes both before and after the fix.

```
tap.test('add success with buffer', function (t) {
  const entry = {
    someAttribute: Buffer.alloc(20)
  }
  t.context.client.add('cn=add, ' + SUFFIX, entry, function (err, res) {
    t.error(err)
    t.ok(res)
    t.equal(res.status, 0)
    t.end()
  })
})
```

jsumners:
> This following test passes, but it passes both before and after the fix.

Then we should investigate why.

UziTech (approved):

jsumners (approved):
Awesome! This is what we need to help improve this library. The test fails before the patch and succeeds after.

jsumners:

when opening a new issue.

PR #764, opened 2021-10-22
Update dn.js

LucidThought:
Change in parse to allow for varying case in the DN provided for searches. This change was made to affect the DN and filters provided for searches, but may have impact that I'm not expecting. Please review


UziTech:
You will have to write tests for this to be merged.

PR #773, opened 2021-12-02
Fix node-ldapjs#329

### LINKED ISSUE #329: Concurrent calls to client.search() failed when using "ldaps"

moghammed:
This PR fixes #329. This is a small improvement over the code changes proposed in the thread of https://github.com/ldapjs/node-ldapjs/issues/329, which break the server.test. 

There is no test included because the error is caused by a race condition and I can't reproduce it reliably.

### DIFF
--- a/lib/client/client.js
+++ b/lib/client/client.js
@@ -1273,11 +1273,18 @@ Client.prototype._sendSocket = function _sendSocket (message,
   log.trace('sending request %j', message.json)
 
   try {
-    return conn.write(message.toBer(), writeCallback)
+    if (expect === 'unbind') {
+      return conn.write(message.toBer(), writeCallback)
+    } else {
+      writeCallback()
+      return conn.write(message.toBer())
+    }
   } catch (e) {
     if (timer) { clearTimeout(timer) }
 
     log.trace({ err: e }, 'Error writing message to socket')
+    emitter.emit('error', e)
+
     return callback(e)
   }
 }


lib/client/client.js
```diff
@@ -1273,11 +1273,18 @@ Client.prototype._sendSocket = function _sendSocket (message,
   log.trace('sending request %j', message.json)
 
   try {
-    return conn.write(message.toBer(), writeCallback)
+    if (expect === 'unbind') {
+      return conn.write(message.toBer(), writeCallback)
+    } else {
+      writeCallback()
+      return conn.write(message.toBer())
+    }
```
jsumners, line 1281: At a minimum we should have a comment here that explains why this code is the way it is. This would at least allow us to accept it without a test, as I know there is difficult code in this project. However, we really should be able to force the condition in our unit tests somehow.

jsumners (changes requested):

jsumners:
👋

Issue #498, opened 2018-12-07
LDAP Connection error


var-kar:
Hello,

I do my connection like this

```
const client = ldap.createClient({
            url: `ldap://url:389`,
            reconnect: true
        });

        client.bind(_this.options.genericEmail, _this.options.genericPassword, err => {
            if (err) {
                callback(err);
            }
        });
```

And unbind after a search.

```
client.unbind( err => {
                        if (err) {
                            console.log(err);
                        }
                    });
```
But after a while on the development server, if it tries to connect again, I get this error and breaks the node server. I am running this application on a kubernetes environment. 

```
events.js:167
--
  | throw er; // Unhandled 'error' event
  | ^
  |  
  | Error: read ECONNRESET
  | at TCP.onStreamRead (internal/stream_base_commons.js:111:27)
  | Emitted 'error' event at:
  | at Socket.onSocketError (/usr/src/app/node_modules/ldapjs/lib/client/client.js:1169:12)
  | at Socket.emit (events.js:182:13)
  | at emitErrorNT (internal/streams/destroy.js:82:8)
  | at emitErrorAndCloseNT (internal/streams/destroy.js:50:3)
  | at process._tickCallback (internal/process/next_tick.js:63:19)
```
Please help me fix this error.

Thank you.

tastypackets:
I'm not 100% sure this is your issue based off what you posted, but looks similar to when the client is left open and connection is reset due to inactivity.

The client is connected the minute it's created, I think it's a poor decision in this lib. If you don't destroy  / disconnect the client you will get an error from the connection timing out. 

The client code is documented saying it can be reused and auto-reconnect, however in my expierence this has not worked so well especially with Active Directory servers.

If the auto-reconnection is working for you then disconnecting the client might be a good option. If not you can create a new client for each bind request.

Barathwaja:
Inside client.bind() fn -> the arguments are bindDN and bindCredentials, instead u were passing **generic email and password.** the first parameter bindDN pass that first.

tastypackets:
You can bind using the UPN, which in many companies is the users email address.

I think this was just an improper choice of words by the lib devs to only mention DN. The if statement just checks if it's a string or an instance of DN, if so it passes it along.

var-kar:
I did try 

`client.destroy();`

But I still get the error on production machines.

Many thanks,
Karthik

tastypackets:
Are you listening for errors on the client? If so can you post that error message?

[...]

PR #786, opened 2022-02-15
Add Note on Client Error Handling to Documentation

FabianScheidt:
Some people (including myself) attempt to use the LDAP client without registering a handler to the _error_ event, since e.g. connection errors are supposed to be included in the callbacks of other methods. Since the client is an `EventEmitter` however the Node.js process will exit if the handler is missing.

This PR adds a note to the documentation to avoid confusion as suggested in #356 and #498.


UziTech (approved):
LGTM

jsumners:

when opening a new issue.

PR #798, opened 2022-03-27
Remove coveralls

jsumners:
This codebase is not well covered to begin with. Plus, we can simply fail from thresholds. Eventually, we should be able to use thresholds of 100% and tools like Coveralls would be completely pointless.


jsumners:

when opening a new issue.

Issue #799, opened 2022-03-29
Replace `lib/messages` with a `@ldapjs/messages` module


jsumners:
This work will be very similar to #797.

jsumners:
https://github.com/ldapjs/messages has been created and I'm working through migrating all of the message type implementations to it. Unit tests in CI are failing at the time of writing because the code depends on an unpublished version of `@ldapjs/asn1`. I do not plan to publish that version until I am certain no further changes need to be made as I implement the remaining message types.

jsumners:
⚠️ This issue has been locked due to age. If you have encountered a recent
problem that seems to be covered by this issue, please open a new issue.

Please include a [minimal reproducible example](https://en.wikipedia.org/wiki/Minimal_reproducible_example)
when opening a new issue.

PR #806, opened 2022-06-05
Replace protocol with @ldapjs/protocol

jsumners:
This is in the same spirit as #797. It lays the ground work for solving #799. All of `client`, `server`, and `messages` use the constants exported by this module.

It should be noted that I changed the names of operations constants from `FOO_REP_BAR` to `FOO_RES_BAR`. I imagine the original naming was intended to be for "reply", but that is not consistent with the spec or even the rest of the code base. People implementing servers with this library **will** need to account for the change.

Note: this PR targets the branch that #797 is based upon.


UziTech (approved):
LGTM

jsumners:

when opening a new issue.

PR #809, opened 2022-06-12
Migrate filter extensions

jsumners:
The purpose of this PR is to migrate the filter extensions to the new `@ldapjs/filter` package and integrate them here. Basically, the original code wrote filters in two parts:

1. A base filter object that represents an LDAP filter string as an object, e.g. `presenceFilter.toString() // (objectclass=*)`
2. A set of extensions to the base objects that facilitate turning them into BER encoded data and decode them from BER encoded data

So, the work here will be to fuse part two into part one, which was previously extracted to `@ldapjs/filter`, and bring that code into `ldapjs` through the module.


UziTech (approved):
LGTM

jsumners:

when opening a new issue.

PR #812, opened 2022-07-04
fix: strict mode problem - setting value in control constructor

### LINKED ISSUE #591: Strict mode problem
Hello there is a problem with this library when using strict mode.

`(node:16368) UnhandledPromiseRejectionWarning: TypeError: Cannot set property value of [object Object] which has only a getter
    at PagedResultsControl.Control (ldapjs\lib\controls\control.js:29:13)       
    at new PagedResultsControl (Cdapjs\lib\controls\paged_results_control.js:31:11)`

I simulated what the library does [here](https://jsfiddle.net/0xdzrpmt/)

Basically in strict mode you cannot set a property that does not have a setter if you use Object.defineProperty.

I can't provide an example usage because I don't use this library directly, but I use another library that has dependence on it, and I can't use my application if I enable strict mode

sastan:
Fixes #591


jsumners (changes requested):
Thank you for the contribution. At a minimum we will need a unit test to cover this. The test should fail prior to the change and pass with the change in place.

However, I'm hesitant to accept any patches to this code path right now. We are very close to v3 (https://github.com/ldapjs/node-ldapjs/projects/2) being ready. In v3, this code lives in https://github.com/ldapjs/controls. If I'm not mistaken, the issue this PR attempts to fix is not present in the new library.

sastan:
Added a test that should fail in non-strict mode. I have added `"use strict"` pragma to the test function but I'm not sure if it would trigger the issue. 

We found this problem after bundling our whole app into a single executable with strict mode.

test/controls/control.test.js
```diff
@@ -51,3 +51,41 @@ test('parse no value', function (t) {
   t.notOk(c.value, null)
   t.end()
 })
+
+test('only value getter in sub-class with strict mode', function (t) {
+  // eslint-disable-next-line quotes
+  "use strict"
```
jsumners, line 57:
```suggestion
  'use strict'
```

test/controls/control.test.js
```diff
@@ -51,3 +51,41 @@ test('parse no value', function (t) {
...
+
+  const assert = require('assert-plus')
```
jsumners, line 59: I'd rather not add any new code using this library. Additionally, we don't need a 100% complete control implementation for the test, just what is necessary to trigger the error.

jsumners (changes requested):

test/controls/control.test.js
sastan: sure. no problem

sastan:
> > I have added `"use strict"` pragma to the test function but I'm not sure if it would trigger the issue.
> 
> Rollback your change (comment it out) in the library code and run your test. If the test fails, it is a good test.

The test wouldn't trigger the issue. It only happens when `'use strict'` is added to lib/controls/control.js. Because there the assignment is made. If I put it there 41 tests are failing.

This issue only happens when ldapjs is bundle with a global strict mode. 

I understand if you don't want to change existing code. We are using patch-package in our app to apply this fix.

If you want to merge it or not we are ok with it. Keep up the great work.

jsumners:
> This issue only happens when ldapjs is bundle with a global strict mode.

We don't support bundling. It's also a very bad practice to force transitive code (i.e. "not your code") into strict mode. The Node.js runtime does not do this for exactly the reason you have discovered: there is lots of code that will not work.

> I understand if you don't want to change existing code. We are using patch-package in our app to apply this fix.

Given that we don't support bundling, and the v2.x branch is soon to be dead, and that it is not possible to write a unit test to induce the issue, I think we will close this PR. Once v3 is released (hopefully by end of July? Don't hold me to it 🥺) I will close out a lot of issues, such as the one this PR attempts to address.

> Keep up the great work.

Thank you. And thank you for trying to resolve the problem with a PR.

jsumners:

when opening a new issue.

PR #825, opened 2022-12-16
Fix minification issue with FilterString

lukaselmer:


lib/client/client.js
```diff
@@ -572,7 +572,7 @@ Client.prototype.search = function search (base,
     options.filter = filters.parseString(options.filter)
   } else if (!options.filter) {
     options.filter = new PresenceFilter({ attribute: 'objectclass' })
-  } else if (Object.prototype.toString.call(options.filter) !== '[object FilterString]') {
+  } else if (!(options.filter instanceof filters.FilterString)) {
```
jsumners, line 575: This will not work. See https://github.com/fastify/fastify-error/pull/86#issuecomment-1301466407 for why.

jsumners (changes requested):

lib/client/client.js
lukaselmer: Agree. And it's handled already by overwriting `toStringTag` (https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/Symbol/toStringTag). Closing this.

jsumners:

when opening a new issue.

PR #829, opened 2023-02-04
Server learned to listens on any address (IPv4 or IPv6) when host is omitted

stalb:
Actually when using `listen` and host is omitted the server accepts connections directed to any IPv4 address but not to IPv6 addresses.

With this patch the server will accept connections on the unspecified IPv6 address (::) when IPv6 is available, or the unspecified IPv4 address (0.0.0.0) otherwise.

Note: In most operating systems, listening to the unspecified IPv6 address (::) may cause the server to also listen on the unspecified IPv4 address (0.0.0.0).

This pull request aligns ldapjs server behavior with that of Net.server.

### DIFF
--- a/docs/server.md
+++ b/docs/server.md
@@ -80,8 +80,11 @@ Example:
 `listen(port, [host], [callback])`
 
 Begin accepting connections on the specified port and host. If the host is
-omitted, the server will accept connections directed to any IPv4 address
-(INADDR\_ANY).
+omitted, the server will accept connections on the unspecified IPv6 address (::)
+when IPv6 is available, or the unspecified IPv4 address (0.0.0.0) otherwise.
+
+In most operating systems, listening to the unspecified IPv6 address (::)
+may cause the net.Server to also listen on the unspecified IPv4 address (0.0.0.0).
 
 This function is asynchronous. The last parameter callback will be called when
 the server has been bound.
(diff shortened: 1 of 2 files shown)


jsumners (changes requested):
Thank you for taking the time to submit an improvement. However, I have a couple of requests:

1. Target the `next` branch. The `master` branch is too diverged from `next`, and I have no intention of releasing something from `master` until `next` has been merged into it.
2. I would rather we change the default to a secure default. That is, instead of defaulting to `0.0.0.0` we should default to `127.0.0.1` or `::1`. The documentation change would indicate that this is the default, and to listen on other addresses they should be specified (and that `0.0.0.0` or `::` can be provided when the user understands the implications).

stalb:
If `localhost` is used as the new default value: 

- this would be a major breaking change (breaking a running app would be less likely with the original proposition)
- the server will behave differently from one version of node to another: in dual stack environment with **node 18** it will listen on `::1` while it will listen on `127.0.01` with **node 16**.

It is quite common to listen on all interfaces by default (for example, this is what _OpenLDAP_ does)

jsumners:
I did not suggest `localhost`. That has other issues. Regardless, the suggestion is a breaking change. The next release will be v3. 

It doesn't matter what OpenLDAP does. I am not willing to default to an insecure value. People should weigh their requirements versus the implications before listening on all interfaces.

stalb:
Ok, I understand this point of view.

PR #834, opened 2023-02-21
Replace messages

### LINKED ISSUE #799: Replace `lib/messages` with a `@ldapjs/messages` module

jsumners:
Resolves #799.

There's a lot of hacks in here to keep the status quo. Working through this has made it abundantly clear that there are way too many side effects in this code, and that the `@ldapjs/client` and `@ldapjs/server` packages/rewrites are a must.

That being said, we are _much_ closer to RFC ("spec") compliance with this PR.


UziTech (approved):
Nice work!

jsumners:
> Nice work!

Thank you. It was _a lot_ 😅

PR #846, opened 2023-03-07
Quick fix for outdated client doc still referring to result.object

Tethik:
Just a minor fix, I'm guessing there is a more correct way to access the attributes.

### DIFF
--- a/docs/client.md
+++ b/docs/client.md
@@ -308,7 +308,7 @@ client.search('o=example', opts, (err, res) => {
     console.log('searchRequest: ', searchRequest.messageId);
   });
   res.on('searchEntry', (entry) => {
-    console.log('entry: ' + JSON.stringify(entry.object));
+    console.log('entry: ' + JSON.stringify(entry.pojo));
   });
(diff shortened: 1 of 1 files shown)


docs/client.md
```diff
@@ -308,7 +308,7 @@ client.search('o=example', opts, (err, res) => {
     console.log('searchRequest: ', searchRequest.messageId);
   });
   res.on('searchEntry', (entry) => {
-    console.log('entry: ' + JSON.stringify(entry.object));
+    console.log('entry: ' + JSON.stringify(entry.pojo));
```
jsumners, line 311: This looks good according to the example. It's just trying to print out a JSON representation of the full search entry object. If the example were to be reworked to show how to access the attributes, that would be `entry.attributes`.

jsumners (approved):
Thank you for helping out.

docs/client.md
Tethik: Good to know! I'll try it out in my project.

PR #864, opened 2023-03-28
transcontextual safe type checks

axkibe:
In case the supplied function goes into a different context (using vm.SourceTextModule for example) the "instanceof" checks fail.

typeof obj === 'function' and Array.isArray are true among contexts. 

BTW: as far I read isArray has explicitly been created for cross-realm situations.

### DIFF
--- a/lib/server.js
+++ b/lib/server.js
@@ -47,15 +47,15 @@ function mergeFunctionArgs (argv, start, end) {
   const handlers = []
 
   for (let i = start; i < end; i++) {
-    if (argv[i] instanceof Array) {
+    if (Array.isArray(argv[i])) {
       const arr = argv[i]
       for (let j = 0; j < arr.length; j++) {
-        if (!(arr[j] instanceof Function)) {
+        if ((typeof arr[j] !== 'function')) {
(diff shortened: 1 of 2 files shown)


jsumners (commented):
This change makes sense. Using `instanceof` is rarely a good idea. How can we add a unit test to guard against regression?

axkibe:
An extremely simplified minimal example of what I was doing. Also switched to vm.runInContext since vm.SourceTextModule is still behind an experimental flag:

```
const ldap = require('ldapjs');
const vm = require('node:vm');
const ctx = vm.createContext( {} );
vm.runInContext(`
globalThis.search=function(){};
globalThis.searches=[function(){}];
`, ctx );
const server = ldap.createServer( );
server.search( '', ctx.search );
server.search( '', ctx.searches );
```
This will fail with instanceof but work with typeof and isArray.

jsumners:
Can you please add a unit test, or integration, that will fail without this change?

axkibe:
Ok, done.

jsumners (approved):
Looks good to me.

jsumners:
https://github.com/ldapjs/node-ldapjs/releases/tag/v3.0.2

Issue #890, opened 2023-05-25
TypeError: Cannot destructure property 'message' of 'tracker.fetch(...)' as it is null


BryanHunt:
ldapjs 3.0.2

I'm getting an occasional crash with the following error.  I do have an error handler registered with the client, so I think this is a case that is not covered.

```
/app/node_modules/ldapjs/lib/client/client.js:884
      const { message: trackedMessage, callback } = tracker.fetch(message.messageId)
                       ^
TypeError: Cannot destructure property 'message' of 'tracker.fetch(...)' as it is null.
    at Parser.onMessage (/app/node_modules/ldapjs/lib/client/client.js:884:24)
    at Parser.emit (node:events:513:28)
    at Parser.write (/app/node_modules/ldapjs/lib/messages/parser.js:135:8)
    at TLSSocket.onData (/app/node_modules/ldapjs/lib/client/client.js:875:22)
    at TLSSocket.emit (node:events:513:28)
    at addChunk (node:internal/streams/readable:315:12)
    at readableAddChunk (node:internal/streams/readable:289:9)
    at TLSSocket.Readable.push (node:internal/streams/readable:228:10)
    at TLSWrap.onStreamRead (node:internal/stream_base_commons:190:23)
    at TLSWrap.callbackTrampoline (node:internal/async_hooks:130:17)
```

jsumners:
A minimal reproduction will be needed to provide support on this.

BryanHunt:
This problem occurs maybe once a day out of 5 running instances.  Reproduction is going to be impossible.  My recommendation is to add a check for the case described in the error message and emit an error back to the client.  That way, I can handle the error in my code rather than the application crash.

BryanHunt:
So you are fine with applications crashing that use your code?

jsumners:
I asked for more information. You declined to give it. I cannot diagnose a problem without a reproduction. There should be zero instances that a message is unavailable. So there is most likely some issue in whatever ldap conversation that resulted in the report.

I am quite willing to collaborate on a diagnosis and solution if you are, but we'll need a reproduction.

janiago:
I have met the same issue in my production ENV. I can try to catch it as "uncaughtException" on process level, but it's not a long term solution. Hope the issue can be solved, otherwise we have to change our LDAP solution.
Thanks!

benzhuo:
In index.js  line 91
[...]

PR #901, opened 2023-06-16
fix the issue 890 which throw exception Cannot destructure property '…

benzhuo:
…message' of 'tracker.fetch(...)' as it is null


jsumners:
Thank you for contributing. We will need a test that fails as described in #890 without this change. My intention is to write an issue specific test that uses `ldapjs.createServer` to create a "server" that will induce the problem in the client. It would be similar to the test https://github.com/ldapjs/node-ldapjs/blob/f2890088e49c0c7b7b880998d73d6e4a448d7b4e/test/issue-845.test.js. I just haven't been able to dedicate the time to it yet.

benzhuo:
I am not very understanding,  this is a very explicit syntax error.

jsumners:
We need to include a unit test that covers the issue in order to guard against regression. I have provided guidance on how I intend to approach the test. I can provide further guidance if you have specific questions, or you can wait until I have time to get to it myself.

test/issue-890.test.js
```diff
@@ -0,0 +1,69 @@
...
+  return new Promise((resolve, reject) => {
+    t.context.client.destroy()
+    server.close((err) => {
+      if (err) return reject(err)
+      resolve()
+    })
+  })
+})
+
+tap.test('tracker.fetch', t => {
```
jsumners, line 61: We can set a timeout for the test like `tap.test('handle null messages', { timeout: 2000 }, t => {`. Then use a lower `setTimeout` in the server to delay the response past the time the connecting client would expect it.

At least, that's my idea for how to write this test.

jsumners:
Thank you for working on this @benzhuo. Unfortunately, the included test does not cover the crashing case. The primary work in this PR is developing that test. So I am closing this PR in favor of #913 which includes a working test.

PR #910, opened 2023-06-27
update modification object in ldap.change

### LINKED ISSUE #865: replace documentation doesnt match with actual behaviour in 3.0.2
Reading the documentation, I would assume this works (it works fine for version 2).

```javascript
const change = new ldap.Change({
  operation: 'replace',
  modification: {
    pets: ['cat', 'dog']
  }
});
```

The defined interface `Change` also suggest that this should work.

```javascript
export interface Change {
    operation: string;
    modification: {
        [key: string]: any;
    };
}
```

What's acutally needed in order to pass the `Attribute.isAttribute` check (if not done like this, an error of `modification must be an Attribute` will be thrown).

```javascript
const input: ldapjs.Change = {
	operation: 'replace',
	modification: {
		type: "pets",
		values: ['cat', 'dog']
	}
};
```

Ps. As a bonus, change the interface to:

```typescript
export interface Change {
    operation: "add" | "delete" | "replace";
    modification: {
        [key: string]: string | number | whatever-is-allowed-here;
    };
}
```

mihir254:
This change resolves #865 by adding updating the modification object on ldap.Change

### DIFF
--- a/docs/client.md
+++ b/docs/client.md
@@ -235,6 +235,16 @@ must be one of:
 | add       | Adds the attribute value(s) referenced in `modification`.  The attribute may or may not already exist. |
 | delete    | Deletes the attribute (and all values) referenced in `modification`. |
 
+```js
+interface Change {
+  operation: 'add' | 'delete' | 'replace',
+  modification: {
+    type: string,
+    values: any
+  }
+}
+```
+
 `modification` is just a plain old JS object with the values you want.
 
 # modifyDN
--- a/docs/client.md
+++ b/docs/client.md
@@ -235,17 +235,13 @@ must be one of:
 | add       | Adds the attribute value(s) referenced in `modification`.  The attribute may or may not already exist. |
 | delete    | Deletes the attribute (and all values) referenced in `modification`. |
 
-```js
-interface Change {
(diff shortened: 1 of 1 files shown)


docs/client.md
```diff
@@ -234,6 +235,16 @@ must be one of:
...
 
+```js
+interface Change {
+  operation: 'add' | 'delete' | 'replace',
+  modification: {
+    type: string,
+    values: any
+  }
+}
+```
```
jsumners, line 246: Let's strike this non-JavaScript block. See my next comment for what we should do instead.

docs/client.md
```diff
@@ -234,6 +235,16 @@ must be one of:
...
+
 `modification` is just a plain old JS object with the values you want.
```
jsumners, line 248: This line should be updated to read similarly to lines 229 and 230 with a similar table following it that shows the properties required.

jsumners (changes requested):
Looking pretty good so far.

docs/client.md
mihir254: Okay, I have added the table instead!

jsumners (approved):
Fantastic! Thank you for the assistance.

PR #912, opened 2023-06-28
Document Updated for filter method

devbysn:
documentation for filter method updated.

### DIFF
--- a/docs/filters.md
+++ b/docs/filters.md
@@ -112,9 +112,9 @@ map to:
 
 ```js
 {
-  initial: 'foo',
-  any: ['bar', 'cat'],
-  final: 'dog'
+  subInitial: 'foo',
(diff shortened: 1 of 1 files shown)


jsumners:
Thank you for taking this on, @devbysn. When I opened the "good first issue" I expected that the true fix would be a while coming (as in, I didn't plan to deprecate the `subX` properties any time soon). But the work has been started by a community member in https://github.com/ldapjs/filter/pull/5 to do just that. Given that PR is progressing, I think that changing this documentation would be unnecessary. What do you think?

devbysn:
Sure, @jsumners .
i will check other issues, will check if i can help in any.

jsumners:
Thanks. I'm going to go ahead and close this, and the issue, as being solved in a different manner.

Issue #919, opened 2023-07-25
Question about accessing search result attributes in v3


ecs-hk:
I'm making the move from ldapjsv2 to v3, and I have read the release notes documentation:
* https://github.com/ldapjs/node-ldapjs/releases/tag/v3.0.0

**My question**: under v3, is there any way to access a search result attribute *directly* (like under v2) so that I don't have to iterate through an array of attributes every time?

Abbreviated example of a search result, `entry.object`, from v2:
```javascript
{
    "dn": "cn=foo,dc=org",
    "member": [
      "uid=baz",
      "uid=fuz"
    ],
    "email": "foo@org",
    "private": "FALSE"
}
```

By contrast, abbreviated example of a search result, `entry.pojo`, from v3:
```javascript
{
  "type": "SearchResultEntry",
  "objectName": "cn=foo,dc=org",
  "attributes": [
    {
      "type": "member",
      "values": [
        "uid=baz",
        "uid=fuz"
      ]
    },
    {
      "type": "email",
      "values": [
        "foo@org"
      ]
    },
    {
      "type": "private",
      "values": [
        "FALSE"
      ]
    }
}
```

Thanks for any help.

jsumners:
I think the comments in issue #841, which is linked from the v3 release notes, are pretty clear. But, to reiterate, the v2 `.object` property was _not_ a direct access. It converted object to an object literal, i.e. serialized it to "JSON", and returned that object. So it was an expensive operation. This is exactly the way the `.pojo` accessor works.

We would definitely welcome a PR to `@ldapjs/messages` to provide utility accessors for the `SearchEntry` object. But as it stands, the best, most efficient, way to access the properties on this object are through the functional methods described in the above referenced issue.

Issue #841, opened 2023-03-01
entry received on a searchEntry message does not contain object


VirgileD:
Hello, I'm updating a project and switch to 3.0.0.
I had to change the code (obviously)  from
```
client.search('cn=groups,dc=rd,dc=test,dc=company,dc=fr', { filter: `(&(memberuid=USERID`, scope: "sub", attributes: "cn"}, function(err, res) {
   res.on('searchEntry', function(entry) {
   groups.push(entry.object.cn)
}});
```
to
```
client.search('cn=groups,dc=rd,dc=test,dc=company,dc=fr', { filter: `(&(memberuid=USERID`, scope: "sub", attributes: "cn"}, function(err, res) {
   res.on('searchEntry', function(entry) {
   groups = groups.concat(entry.pojo.attributes.filter(f => f.type=="cn").map(m => m.values)[0])
}});
```
Obviously the "object" field of the returned entry is gone and I didn't find anything else but the "pojo" field to retrieve the results of the search.
Not a big problem though but it took me some time to figure it out and the documentation should be changed accordingly unless I missed something.

jsumners:
Yep. That's a change in the way the message objects are constructed and accessed. If you want a "plain JavaScript" object of a message then `.pojo` is the accessor you're looking for.

However, in this specific case, I would skip `.pojo` and do:

```js
groups = groups.concat(
  entry.attributes.filter(f => f.type === 'cn').map(m -> m.values).pop()
)
```

Thank you for highlighting this change. I have updated the release notes to reflect this item https://github.com/ldapjs/node-ldapjs/releases/tag/v3.0.0.

VirgileD:
ah, ok, I had missed the entry.attributes part, makes sense. Thank you!

jsumners:
Understandable. We need some prominent docs around everything. Unfortunately, the jsdoc gets lost in the spaghetti that is the client/server code. It would be more evident when using the libraries directly. Hopefully this all gets sorted out in time, but will be a bit difficult for quite some time.

mil7:
Hi there,
is there any plan to provide a more convenient access to the entry's attributes?

I am curious, what is the reason for this change? We heavily depended on the v2's object representation of the attributes. Maybe we did not understand LDAP well enough.

jsumners:
> is there any plan to provide a more convenient access to the entry's attributes?

What is more convenient than `entry.attributes`?

> I am curious, what is the reason for this change?

[...]

PR #920, opened 2023-07-26
Provide accessor for SearchEntry object

ecs-hk:
Proposing this small change as a more intuitive way to get attribute values under v3.

Related to my question about, and the response to, issue #919. The code change borrows ideas from #841.

### DIFF
--- a/lib/messages/search_response.js
+++ b/lib/messages/search_response.js
@@ -74,7 +74,10 @@ SearchResponse.prototype.send = function (entry, nofiltering) {
     entry = new SearchEntry({
       objectName: typeof (save.dn) === 'string' ? parseDN(save.dn) : save.dn,
       messageId: self.messageId,
-      attributes: Attribute.fromObject(entry.attributes)
+      attributes: Attribute.fromObject(entry.attributes),
+      getAttribute: function(a) {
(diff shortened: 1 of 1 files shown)


jsumners (changes requested):
Thank you for working on this. However, it should be added https://github.com/ldapjs/messages and it must include unit tests.

I think the approach is going in the right direction. It's similar to the direction I thought of overnight. I think we could do this is one of two, or both, ways:

1. When the attributes are added, decorate the `this.attributes` object with getters, e.g. something like:
```js
for (const attr of attributes) {
  Object.defineProperty(this.attributes, attr.name, {
    get () {
      return attr.value
    }
  })
}
```

2. Define a method on `SearchEntry.prototype.attributes` and accepts an attribute name to get. This would be the scenario you have defined.

jsumners:
> It may take me a bit to understand the testing for this project, and write appropriate unit tests, but I will try.

Do your best and we will sort through any changes needed during review.

I suggest looking at https://github.com/ldapjs/messages/blob/041490fb3a33634b8457c06dd75985daea39ff3c/lib/messages/search-result-entry.js#L85-L101

It may be worth creating an `attributes` prototype object that extends `Array`, as right now it is implemented with a simple array literal.

